PR #11, opened 2015-01-06
Use string IDs for models maintained in PM memory

kraman:
Ensures that there are no ID conflicts when collecting data from multiple PMs and processing them in mesh API. Also allows for consistent access to models related to the Instance.

* Change Action, ProfileData, ServiceProcess models to have string ID
* Create String IDs instead of using autogenerated numeric IDs
* Make foreign keys to ServiceInsance consistent (serviceInstanceId)
* Add relation between ServiceProcess and ProfileData
* Add relation between ServiceMetric and ServiceInstance

### DIFF
--- a/models/action.json
+++ b/models/action.json
@@ -3,6 +3,10 @@
   "base": "PersistedModel",
   "description": "Represents an action to perform on an instance",
   "properties": {
+    "id": {
+      "type": "string",
+      "id": true
+    },
     "request": {
       "type": "object",
       "required": true,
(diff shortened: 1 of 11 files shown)


sam-github:
This isn't backwards compatible, is it? Shouldn't the package.version needs to go to `4.0.0`, and the minimum dependency in https://github.com/strongloop/strong-pm/pull/94 should go to `^4`?

Is this going to effect arc? It looks like the changes may not, but I am not certain from just eyeballing the diff.

kraman:
Yes, it is not backward compatible. Will increment the major version. Arc will need to change to comply with this new structure. They currently have a few hard-coded IDs which need to change.

/cc @chandadharap

models/action.json
```diff
@@ -15,11 +19,16 @@
     "result": {
       "type": "object",
       "description": "The result or output of the action"
+    },
+    "serviceInstanceId": {
+      "type": "number",
```
rmg: It is a shame these foreign keys have to be defined just to mark them as required instead of being able to define the relation as required instead.

Fun fact: the `type`'s for these foreign keys appear to get overridden by the type specified on the id field when the relation is defined.

PR #13, opened 2015-01-20
discussion/spike: Model update proposal

kraman:
Summary of changes:
* Separate ServiceInstance into ServiceContainer (stored in redis) and ServiceInstance (stored in PM-memory)
* Hide update/delete methods for read-only models
* Rename ServerService -> Service
* Rename InstanceAction -> Action
* Convert IDs to be GUID strings instead of numeric

Unimplemented ideas:
* Move to remote-methods like /instance/stop?soft=true instead of a soft-stop Action


models/metric.txt
```diff
@@ -1,66 +0,0 @@
-service/server-service:
```
sam-github: should be a seperate commit, its unrelated

models/executor.js
```diff
@@ -5,4 +5,28 @@ module.exports = function(Executor) {
     }
     next();
   };
+
+  Executor.disableRemoteMethod('__delete__instances');
```
sam-github: why?

sam-github: I mean, why are any of these being disabled? We have REST API, they use it, what problem are we solving?

models/metric.txt
sam-github: also, this file was my attempt at understanding the relationships between the models

IMO, @altsang @ritch @seanbrookes @raymondfeng Loopback/strong-arc ***desperately*** needs a model relationship viewer that graphically displays the relationships between models. Figuring it out from the json is pretty hard, I always end up drawing a picture on paper, or writing a text file summary like this.

models/profile-data.js
```diff
@@ -8,4 +8,10 @@ module.exports = function(ProfileData) {
     this.id = this.id || profileId;
     next();
   };
+
+  ProfileData.disableRemoteMethod('create', true);
```
sam-github: I think code like this deserves a comment.

models/executor.js
kraman: Some of these models are managed and meant to be created/deleted only by PM or Mesh etc. They should only be queryable and not editable over the REST API.

models/metric.txt
kraman: I'll split this out add a SVG diagram to the repo to explain

models/profile-data.json
```diff
@@ -6,15 +6,16 @@
   "properties": {
     "id": {
       "type": "string",
-      "index": true,
-      "id": true
+      "required": true,
```
sam-github: I think only required changed, but the order of the other two fields was switched.

models/service-container.js
```diff
@@ -0,0 +1,7 @@
+module.exports = function(ServiceContainer) {
```
sam-github: ?

sam-github:
I think this would be easier to read if each change was an individual commit, like your point form list in the PR. Its quite difficult to read large changes to json, particularly because its not possible to put comments in json

Other than that, LGTM, though I've concerns about backwards compatibility.

PR #14, opened 2015-01-26
Add GUIDs and accessToken support

kraman:
Connect to strongloop-internal/scrum-nodeops#123

### DIFF
--- a/models/executor.js
+++ b/models/executor.js
@@ -1,8 +1,14 @@
+var uuid = require('uuid');
+
 module.exports = function(Executor) {
-  Executor.beforeCreate = function(next) {
-    if (this.remainingCapacity === -1) {
+  Executor.definition.properties.id.default = function() {
+    return uuid.v4();
+  };
+
+  Executor.beforeSave = function(callback) {
+    if (this.remainingCapacity === null && this.totalCapacity !== null) {
       this.remainingCapacity = this.totalCapacity;
     }
-    next();
+    process.nextTick(callback);
   };
 };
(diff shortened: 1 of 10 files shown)


models/executor.js
```diff
@@ -1,8 +1,14 @@
...
+
 module.exports = function(Executor) {
-  Executor.beforeCreate = function(next) {
-    if (this.remainingCapacity === -1) {
+  Executor.definition.properties.id.default = function() {
+    return uuid.v4();
+  };
+
+  Executor.beforeSave = function(callback) {
+    if (this.remainingCapacity === null && this.totalCapacity !== null) {
```
rmg: This looks like initialization code, but it is in `beforeSave` instead of `beforeCreate`. Is this in place of a implementing `remainingCapacity.default()`?

rmg:
UUID4/GUID seems like it would be an excellent type for loopback to support "natively". Should we file that as a feature request?

Are large numbers of instances going to be created (or instantiated and then GC'd) _en masse_? Do we need to worry about uuid generation overhead? I'm guessing not, just wanted to mention it.

rmg:
@kraman couple questions, but otherwise LGTM.

models/executor.js
kraman: @rmg Moved it to beforeSave because the steps of registering an executor have changed. You must first create the executor and then start it with the generated ID and auth token. Once it starts, it updates the values.

Cant use `default` function since that needs to be independent and cannot be based on values of other fields in the model.

kraman:
@rmg I have already put in a request for loopback to support GUIDs for IDs. They are working on it and it should be available in a few releases from now.

PR #15, opened 2015-01-27
Mesh-model changes for gateway integration

kraman:
Connected to strongloop/strong-scheduler#14


sam-github:
I like red/green/refactor, but I like to fixup before showing the result to people.

PR #16, opened 2015-02-04
Mesh-model changes for gateway integration

kraman:
Connected to strongloop/strong-scheduler#14


sam-github:
A couple questions, and a nit or two, but LGTM.

And this was a ***very*** readable PR, super easy to see what the changes were, and why they were being made, which is challenging with the loopback JSON, thanks for taking the time to present it so well!

kraman:
Merged into branch 4.x

PR #17, opened 2015-02-18
Add support for Express Usage records

kraman:
When the supervised express-base application has strong-express-metrics configured, strong-agent and strong-supervisor will forward usage records which will be stored in the ExpressUsageRecord model.

Connect to strongloop/strong-pm#125

### DIFF
--- a/models/express-usage-record.json
+++ b/models/express-usage-record.json
@@ -15,16 +15,7 @@
       "type": "date",
       "required": true
     },
-    "client": {
-      "type": "object"
-    },
-    "request": {
-      "type": "object"
-    },
-    "response": {
-      "type": "object"
-    },
-    "data": {
+    "usage": {
       "type": "object"
     }
   },
--- a/models/express-usage-record.json
+++ b/models/express-usage-record.json
@@ -1,0 +1,35 @@
+{
+  "name": "ExpressUsageRecord",
+  "base": "PersistedModel",
+  "properties": {
+    "processId": {
+      "type": "number",
+      "description": "id of ServiceProcess, not the pid",
+      "required": true
+    },
+    "workerId": {
+      "type": "number",
+      "required": true
+    },
+    "timestamp": {
+      "type": "date",
+      "required": true
+    },
+    "client": {
+      "type": "object"
+    },
+    "request": {
+      "type": "object"
+    },
+    "response": {
+      "type": "object"
+    },
+    "data": {
+      "type": "object"
+    }
+  },
+  "validations": [],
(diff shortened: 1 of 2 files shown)


kraman:
Connect to strongloop/strong-pm#123

kraman:
Connect to strongloop/strong-pm#125

bajtos:
> When the supervised express-base application has strong-express-metrics configured, strong-agent and strong-supervisor will forward usage records which will be stored in the ExpressUsageRecord model.

For the record, the first two parts are already implemented: strong-agent and strong-supervisor are already forwarding express usage records.

bajtos:
@kraman why is this issue assigned to me? What would you like me to do? Since strongloop/strong-pm#125 is on your plate, I don't see any work left for me.

sam-github:
@bajtos for review.

In particular: https://github.com/strongloop/strong-mesh-models/blob/feature/express-usage/models/express-usage-record.json#L18-L29

Will there only ever be these four top-level properties in the express record?

Its not clear to me how flexible the record is intended to be, or who is to consume it. The idea here is that the gateway is going to emit these customized records, right? And then some arc component is going to consume them for display?

bajtos:
Oh, somehow I did not realise this is a pull request. Makes sense now :)

> Will there only ever be these four top-level properties in the express record?

See the discussion in strong-pm, most notably  https://github.com/strongloop/strong-pm/pull/123#issuecomment-75106363

If we agree that only those four top-level properties is the contract, then it will be like that.

> Its not clear to me how flexible the record is intended to be, or who is to consume it. 

> The idea here is that the gateway is going to emit these customized records, right?

Yes.

> And then some arc component is going to consume them for display?

Yes.

We don't know how flexible the record should be, because we don't have the full vertical slice implemented. Once we start working on the GUI part and also the integration with the Gateway, we will know more about the data we need to collect and process.

For now, I think it's ok to go with the current solution.

bajtos:
The patch LGTM.

sam-github:
LGTM

models/express-usage-record.json
```diff
@@ -0,0 +1,26 @@
...
+    },
+    "workerId": {
+      "type": "number",
+      "required": true
+    },
+    "timestamp": {
+      "type": "date",
+      "required": true
+    },
+    "usage": {
```
bajtos: Can we rename this to `details` please?

kraman: @bajtos Sure. I'll do that in a follow on issue/PR

PR #19, opened 2015-02-24
Move REST API implementation into strong-mesh-models

kraman:
Connect to strongloop/strong-pm#129


package.json
```diff
@@ -1,7 +1,6 @@
 {
-  "name": "strong-mesh-models",
-  "version": "3.0.1",
-  "description": "Loopback models for communicating over REST with strong-pm and strong-scheduler",
+  "name": "strong-mesh-core",
```
rmg: Changing the package name confuses CI. It either needs to match the repo name or be set explicitly in the job config.

server/datasources.json
```diff
@@ -0,0 +1,10 @@
+{
+  "memory": {
```
rmg: Can we take this PR as an opportunity to switch to intention/purpose based names instead of implementation based names?

server/datasources.local.js
```diff
@@ -0,0 +1,22 @@
+var debug = require('debug')('MeshModels.server.datasource');
+var url = require('url');
+
+var dbUrl = url.parse(process.env.STRONGLOOP_MESH_DB || 'memory://');
```
rmg: Would this ever be something other than a "DB", like a web service of some kind?

kraman: We might switch to a SL Db instead of Redis/Memory but will probably never back it with a web service.

server/datasources.json
kraman: `memory` in this case is really intended to be using memory to enable quick lookup of ACLs etc.
The other datasource is called `db` back by memory by default and configurable to use other DBs

kraman:
mesh-models will be used on both server and client. I decided not to move the Git/tar receiving code into mesh-models in order to not have cicada etc deps included in the client.

It still makes sense to move the receive hooks into a (new) module but we can do that later when the scheduler is being refactored.

package.json
```diff
@@ -15,20 +15,32 @@
...
   "scripts": {
-    "pretest": "jshint . && eslint --ignore-path .jshintignore **/*.js",
-    "test": "tap --tap --stdout --stderr --timeout=300 test/cover.js",
-    "cover": "istanbul cover --print tap --tap --stdout --stderr --timeout=300 test/cover.js"
+    "eslint": "eslint client common server test",
+    "jscs": "jscs client common server test",
+    "test": "tap --tap --stdout --stderr --timeout=300 test/test-*.js",
+    "cover": "test/cover.sh"
+  },
+  "dependencies": {
```
sam-github: this is a mix of dev and non-dev dependencies (eslint)

package.json
```diff
@@ -15,20 +15,32 @@
...
+    "jscs": "^1.11.3",
+    "lodash": "^3.3.1",
+    "loopback": "^2.12.0",
+    "loopback-boot": "^2.6.0",
+    "loopback-datasource-juggler": "^2.18.0",
+    "request": "^2.53.0",
+    "serve-favicon": "^2.2.0",
+    "tap": "^0.6.0"
+  },
+  "optionalDependencies": {
```
sam-github: never been sure what the point of making this optional is... npm installs it always.

client/client.js
```diff
@@ -0,0 +1,169 @@
...
+
+  var client = loopback();
+  client.dataSource('remote', {'connector': 'remote', 'url': this.apiUrl});
+  boot(client, __dirname);
+
+  this.loopback = client;
+  this.models = client.models;
+
+  // Resolve the models
+  client.models();
```
sam-github: why do we create the method on this, and then also call it right away? is it to populate the cache, or something?

client/client.js
```diff
@@ -0,0 +1,169 @@
+var assert = require('assert');
+var _ = require('lodash');
+var boot = require('loopback-boot');
+var debug = require('debug')('MeshModels.client');
```
sam-github: odd debug naming. is MeshModels is supposed to be a more readable strong-mesh-models?

and shouldn't the `.` be a `:`?

I think I'd stick with `strong-mesh-models:client`, it might be verbose, but it is very predictable, I think most of our code uses that pattern.

client/client.js
```diff
@@ -0,0 +1,169 @@
+var assert = require('assert');
+var _ = require('lodash');
```
sam-github: not sorted lexically

client/config.json
```diff
@@ -0,0 +1,2 @@
+{
```
sam-github: mandatory, I gather?

client/datasources.json
```diff
@@ -0,0 +1,6 @@
+{
```
sam-github: does client need a datasources? I'd think its all remote, so wouldn't.

client/model-config.json
```diff
@@ -0,0 +1,68 @@
+{
```
sam-github: why are some things remote and some in-memory?

I wish loopback would switch to an extended json format that allowed comments... @bajtos @raymondfeng yaml, json5, anything (while still supporting vanilla json)

package.json
rmg: Doesn't make much sense in this case, but it can help with binary deps that have pure js fallbacks as separate modules.

server/datasources.local.js
```diff
@@ -0,0 +1,22 @@
+var debug = require('debug')('MeshModels.server.datasource');
```
sam-github: `:server:datasource`? see other comments.

server/model-config.json
```diff
@@ -0,0 +1,69 @@
+{
+  "_meta": {
+    "sources": [
+      "loopback/common/models",
+      "loopback/server/models",
+      "../common/models",
+      "./models"
```
sam-github: We would never have both ./common/models and ./models, would we?

server/server.js
```diff
@@ -0,0 +1,41 @@
...
+    this._server.close(function() {
+      app.emit('stopped');
+      if (callback) callback();
+    });
+  };
+
+  return app;
+}
+
+// start the server if `$ node server.js`
```
sam-github: Schoon started this, intending that all servers be composable since you could require them and add them as handlers, as well as be runnable. Its a good idea in terms of goal, but I don't think the mechanics of this convention are working for us.

We can fix after this PR, but essentially, this mixes require of a module and run of a module in one file... attempting to deduce what was intended. However, the deduction heuristics are just heuristics, and fail regularly. I think it makes more sense for the package, if runnable, to have a top-level `server.js` (this is what `npm start` looks for - in the absence of a start script - and also what `slc run` looks for) and whatever is exported by the package (by package `main:`, or top-level index.js) should never run code, and return a constructor for the app. Perhaps two... a "full ctor" (including the err handler, etc.), and a "minimal" that is *only* the api's exposed by this app' models. That probably means we shouldn't call boot(), but instead break it into two stages: minmal (no middleware), and full-app (with all the middleware).

You've already run into some issues with this, lets talk sometime, and work through what's reasonable.

server/server.js
```diff
@@ -0,0 +1,41 @@
+/* eslint no-console:0 */
+var loopback = require('loopback');
+var boot = require('loopback-boot');
+var ServiceManager = require('./service-manager');
```
sam-github: ServiceManager should be exported.

server/boot/authentication.js
```diff
@@ -0,0 +1,4 @@
+module.exports = function enableAuthentication(server) {
+  // enable authentication
```
sam-github: LB boilerplate, right? The comment is text book "don't write comments like these"... it's just the method name, converted from "camelcase" to "space separated" (with acronym expanded) :-(

server/boot/extend-models.js
```diff
@@ -0,0 +1,5 @@
+var serviceInstance = require('../models/service-instance');
+
+module.exports = function(server) {
+  serviceInstance(server.models.ServiceInstance);
```
sam-github: This is the thing LB is going to do for us, in an upcoming revision, right?

server/models/service-instance.js
```diff
@@ -0,0 +1,70 @@
+var async = require('async');
+var debug = require('debug')('MeshModels.server.ServiceInstance');
+
+module.exports = function(ServiceInstance) {
+  ServiceInstance.beforeRemote(
```
sam-github: method beforeRemote

server/models/service-instance.js
```diff
@@ -0,0 +1,70 @@
...
+      debug('updateAttributes with %j', ctx.args.data);
+      // When updating the instance via REST, only allow changes to cpus
+      ctx.args.data = {
+        cpus: ctx.args.data.cpus
+      };
+      next();
+    }
+  );
+
+  ServiceInstance.observe('before save', function beforeUpdate(ctx, next) {
```
sam-github: observe string "before save"... what's going on? does LB have two generations of ways of supplying hooks? can we use one way, or do we have to use the different ways?

server/models/service-instance.js
```diff
@@ -0,0 +1,70 @@
...
+
+  ServiceInstance.observe('before delete', function(ctx, next) {
+    var serviceManager = ServiceInstance.app.serviceManager;
+
+    ctx.Model.find(ctx.where, function(err, instances) {
+      if (err) next(err);
+      async.each(
+        instances,
+        function(instance, callback) {
+          serviceManager.onInstanceDestroy(instance, callback);
```
sam-github: for save, manager gets notified after... for delete, it gets notified before... why? probably there is a reason for the asymetry.

server/boot/model-explorer.js
```diff
@@ -0,0 +1,26 @@
+module.exports = function mountLoopBackExplorer(server) {
+  var explorer;
+  try {
+    explorer = require('loopback-explorer');
+    var restApiRoot = server.get('restApiRoot');
+
+    var explorerApp = explorer(server, {basePath: restApiRoot});
+    server.use('/explorer', explorerApp);
```
sam-github: hm... this isn't going to be the right location, is it?

I wonder if `smm = require('strong-mesh-models')` should expose:

- app() (no middleware, route starts at `/`, so pm can mount where it wants... maybe `/v1`, maybe not)
- server() (middleware, explorer, and start/stop methods)
- explorer() (this...)

so that the explorer can be mounted where it makes sense to the encapsulating app...

client/client.js
kraman: this is a loopback issue .. `client.models()` populates the cache of models so you can access them with `client.models.ModelName`

client/config.json
kraman: Not mandatory but is LB boilerplate. I opted to stay as close to LB generated structure as possible.

server/boot/model-explorer.js
sam-github: - app() looks like it might be `loopback().rest()`, based on server/boot/rest-api.js

client/model-config.json
kraman: Role and ACL related models are in memory since they are used mostly for caching credentials on the client. They also wont be exposed on the server so cant be accessed over the remote connector.

server/boot/rest-api.js
```diff
@@ -0,0 +1,4 @@
+module.exports = function mountRestApi(server) {
```
sam-github: this is the core REST API, everything else might not be needed by a parent app

client/datasources.json
kraman: client needs the memory datastore for maintaining cache of access tokens etc. see comment below.

server/boot/root.js
```diff
@@ -0,0 +1,6 @@
+module.exports = function(server) {
```
sam-github: like this... not needed when not running stand-alone

server/model-config.json
kraman: We do have both currently since we have `Service` and `Action` in `common/models`, extended as `ServerService` and `InstanceAction` in the `./models`. I had fixed this on the 3.x PM and 4.x mesh-models branch to only use `Service` and `Action` without having to extend them. However that requires a API rev so I kept it out of this PR.

server/server.js
kraman: I export it in a later commit.

server/boot/authentication.js
kraman: Yes, LB boilerplate

server/boot/extend-models.js
kraman: Yes, that is in their backlog.

server/models/service-instance.js
kraman: Yes LB moved away from `beforeSave(...)` style to the `observe('before save', function...` style. I modified all the models to follow the new style over a few different commits

server/models/instance-action.js
```diff
@@ -0,0 +1,124 @@
...
+      if (err) return callback(err);
+
+      var pathname = [
+        InstanceAction.app.get('restApiRoot'),
+        ServerService.sharedClass.http.path,
+        service.id,
+        ProfileData.sharedClass.http.path,
+        String(profile.id),
+        'download'
+      ].join('/').replace(/\/+/g, '/'); // Compress // to /
```
sam-github: http://nodejs.org/api/path.html#path_path_normalize_p might do this? or might not... I guess it might look for `\\` on win32

server/models/service-instance.js
kraman: This is intentional. For save, the manager is notified after the model has been persisted in DB so that queries on the DB will return correct information. Similarly for delete, the manager  is notified before the model has been deleted from the DB, so that queries will return information.

server/models/service-instance.js
kraman: Not sure what you mean here...

server/models/server-service.js
```diff
@@ -0,0 +1,113 @@
...
+      // Full save of Service
+      var service = ctx.instance;
+      if (service._groups.length === 0) {
+        service._groups.push(new Group({id: 1, name: 'default', scale: 1}));
+      }
+      if (service._groups.length === 1 && service._groups[0].scale <= 0) {
+        service._groups[0].scale = 1;
+      }
+      process.nextTick(next);
+    }
```
sam-github: don't need a multi-service branch?

server/boot/model-explorer.js
kraman: I still need to clarify with @ritch on this. Im not sure what happens to /explorer and if it supports exposing models from sub-components of the parent app.

server/server.js
kraman: Sure, lets discuss after this PR.

server/boot/root.js
kraman: Yah, I'll remove this and the parent app can add it if needed

server/models/instance-action.js
kraman: yes, that was my concern. Do you know of any node function that normalizes HTTP paths? perhaps I would url.parse + url.format ?

server/service-manager.js
```diff
@@ -0,0 +1,49 @@
+var debug = require('debug')('MeshModels.server.ServiceManager');
+
+function ServiceManager() {
+  debug('ServiceManager created');
+}
+
+function onServiceUpdate(service, callback) {
```
sam-github: Will we want to know the diff between create and update? Can we know?

server/service-manager.js
```diff
@@ -0,0 +1,49 @@
...
+  debug('onServiceUpdate(%j)', service);
+  setImmediate(callback);
+}
+ServiceManager.prototype.onServiceUpdate = onServiceUpdate;
+
+function onServiceDestroy(service, callback) {
+  debug('onServiceDestroy(%j)', service);
+  setImmediate(callback);
+}
+ServiceManager.prototype.onServiceDestroy = onServiceDestroy;
```
sam-github: Should Destroy be Delete, so it parallels the rest API more closely, or is it better to use a different name to abstract away?

server/models/server-service.js
kraman: This is only setting default groups for a new service that is created without groups. AFAIK the multi service case is only hit when updating multiple models, not creating new ones.

server/service-manager.js
kraman: LB does not provide any way to know the difference.

server/models/service-instance.js
```diff
@@ -0,0 +1,70 @@
...
+        ctx.instance.cpus = 'CPU';
+      }
+    } else if (ctx.data && ctx.data.cpus &&
+      isNaN(parseInt(ctx.data.cpus, 10))) {
+      ctx.data.cpus = 'CPU';
+    }
+    next();
+  });
+
+  ServiceInstance.observe('after save', function(ctx, next) {
```
sam-github: This code is pretty much cut-n-paste, and verbose: I see a copy in at least server-service.js, I think it would be better refactored into some common place. That would also allow it to be clear that the code is actually the same... peering closely, I see some minor differences (this version is shorter by pulling serviceManager out as a var), but I think its otherwise the same.

server/service-manager.js
kraman: LB aliases delete/destroy/remove etc. I prefer to call it destroy because in the scheduler case, its not just removing the model, but also cleaning up resources

client/client.js
sam-github: `populates the cache of models so you can access them with client.models.ModelName` is the comment I was looking for, much clearer than "Resolve".

client/config.json
sam-github: OK.

client/datasources.json
sam-github: ok

client/model-config.json
sam-github: ok.

server/models/service-instance.js
kraman: Should be in only 2 places. Might be able to pull this out into a boot script but having it in the model itself is still my preference.

package.json
sam-github: Right, optional for things that can fail to npm install is useful, optional for things that will always install sucessfully, though... not so useful.

server/boot/authentication.js
sam-github: ok, then leave alone, I guess :-(

server/boot/model-explorer.js
sam-github: in the pm case... I suspect that it would not itself have any models, so that it would do:

- / -> it's status
- /api -> core API
- /explorer -> core explorer
- /... -> cicada
- * error middleware

oh, but I see your point, explorer might need to be told, at creation time, where it will be mounted, and where the REST api it is exploring will be mounted? I thought the explorer had built-in knowledge of the API it is exploring, but does it fetch it all dynamically via REST when it loads? In which case, the strong-mesh-models module wouldn't need to expose an explorer: pm could instantiate its own if it wanted.

server/boot/root.js
sam-github: By remove, you mean delete from strong-mesh-models? I think it should be here, its useful when we run this repo stand-alone, its just that the REST API exported from strong-mesh-models shouldn't include it. If that's possible...

server/model-config.json
sam-github: Sorry, I was thinking of "./" referring to package root, ignore me.

server/models/instance-action.js
sam-github: not sure... it really, really looks like something a module should take care of. I guess its NBD if it works, but regex URL twiddling makes me nervous... what kind of quoting can it damage? What about malformed URLs, that we remap to a different invalid form, and that then shows up in logs, and nobody can figure out where it is from? Maybe that specifically can't happen. But its working OK now. url.parse/format sounds plausible.

server/models/server-service.js
sam-github: ^-- above would make a good comment

server/models/service-instance.js
sam-github: but not the beforeRemote method a dozen lines up, shouldn't it be using the observer() style?

server/models/service-instance.js
sam-github: ^--- good in-source comment

server/server.js
sam-github: I'm commenting on the complete file diff... (actually, I'm reading in vim, diffs suck for something this large, and then looking for files in the complete PR diff).

All of which is to say: did you not push that change?

server/server.js
sam-github: sounds good

sam-github:
Done review, looks like its coming along really well.

server/models/service-instance.js
kraman: @sam-github see http://docs.strongloop.com/display/public/LB/Operation+hooks
beforeRemote was not changed.

server/boot/root.js
kraman: Not sure its possible to make that detection. In that case, I'll just leave it here and if needed the parent app can override it.

package.json
kraman: fixed

server/server.js
kraman: fixed. see https://github.com/strongloop/strong-mesh-models/pull/19/files#diff-168726dbe96b3ce427e7fedce31bb0bc

server/models/service-instance.js
kraman: done

server/boot/model-explorer.js
kraman: left as-is till I can discuss options with Ritchie. Will probably address it as a separate PR if there are changes required.

server/models/instance-action.js
kraman: left as-is for now.

server/models/server-service.js
kraman: done

server/server.js
```diff
@@ -0,0 +1,41 @@
...
+
+function server(serviceManager) {
```
sam-github: This is a constructor, why isn't it capitalized? I understand `loopback()` and `express()` aren't capitalized (for some historical reason, I assume, they are clearly constuctors, not sure what's going on there). I noticed this in the export, where meshServer is exported, which gave me pause: was a server *instance* being exported, or a constructor for a server? The naming implied the former, but I expected the latter. So, I think the naming convention matters.

server/models/service-instance.js
sam-github: Incredibly confusing.

sam-github:
Some small nits, the only major comment is the EEification of the interface base, ServiceManager.

server/server.js
rmg: @sam-github this is more of a factory (like `loopback()` and `express()`). Notice the lack of `this` usage in the body.

kraman: @rmg has the right idea. This is a factory which returns an LB app, not a constructor.

client/model-config.json
rmg: re: extended json format

Are we still getting benefit from using the JSON+loopback-boot combination? Could we replace the static JSON with equivalent JS instead? We already have to define the datasource for the client using the API instead of JSON anyway..

PR #20, opened 2015-03-03
Factor out model updates and move them to mesh-models

kraman:
Connect to strongloop/strong-pm#131

### DIFF
--- a/server/models/agent-trace.js
+++ b/server/models/agent-trace.js
@@ -1,0 +1,43 @@
+var async = require('async');
+var debug = require('debug')('strong-mesh-models:server:agent-trace');
+var deleteOld = require('./delete-old');
+
+module.exports = function extendAgentTrace(AgentTrace) {
+  function recordTrace(instanceId, req, callback) {
+    debug('trace data: %j', req.trace);
+    var ServiceProcess = AgentTrace.app.models.ServiceProcess;
+
+    var q = {where: {
+      serviceInstanceId: instanceId,
+      pid: +req.processId,
+      stopTime: null
+    }};
+
+    function saveTrace(proc, traceData, asyncCb) {
+      var tData = {
+        processId: proc.id,
+        workerId: proc.workerId,
+        timeStamp: traceData[0],
+        trace: traceData
+      };
+      AgentTrace.upsert(tData, asyncCb);
+    }
+
+    return ServiceProcess.findOne(q, function(err, proc) {
+      if (err) return callback(err);
+      if (!proc) {
+        debug('Ignoring trace for unknown process: %d', req.processId);
+        return callback();
+      }
+
+      return async.each(req.trace.list,
+        saveTrace.bind(null, proc),
+        function(err) {
+          if (err) return callback(err);
+          deleteOld(AgentTrace, callback);
+        }
+      );
+    });
+  }
+  AgentTrace.recordTrace = recordTrace;
+};
--- a/server/models/express-usage-record.js
+++ b/server/models/express-usage-record.js
@@ -1,0 +1,38 @@
+var debug = require('debug')('strong-mesh-models:server:express-usage');
+var deleteOld = require('./delete-old');
+
+module.exports = function extendExpressUsageRecord(ExpressUsageRecord) {
+  function recordUsage(instanceId, req, callback) {
+    debug('Express usage record: %j', req.record);
+    var ServiceProcess = ExpressUsageRecord.app.models.ServiceProcess;
+    var record = req.record;
+    var pid = +record.process.pid;
+
+    var q = {where: {
+      serviceInstanceId: instanceId,
+      pid: pid,
+      stopTime: null
+    }};
+
+    ServiceProcess.findOne(q, function(err, proc) {
+      if (err) return callback(err);
+      if (!proc) {
+        debug('Ignoring metrics for unknown worker pid: %d', pid);
+        return callback();
+      }
+
+      var usageRecord = {
+        processId: proc.id,
+        workerId: proc.workerId,
+        timeStamp: record.timeStamp,
+        detail: record
+      };
+
+      ExpressUsageRecord.create(usageRecord, function(err) {
+        if (err) return callback(err);
+        deleteOld(ExpressUsageRecord, callback);
+      });
+    });
+  }
+  ExpressUsageRecord.recordUsage = recordUsage;
+};
--- a/server/models/service-instance.js
+++ b/server/models/service-instance.js
@@ -1,7 +1,7 @@
 var async = require('async');
 var debug = require('debug')('strong-mesh-models:server:service-instance');
 
-module.exports = function(ServiceInstance) {
+module.exports = function extendServiceInstance(ServiceInstance) {
   ServiceInstance.beforeRemote(
     'prototype.updateAttributes',
     function(ctx, _, next) {
@@ -40,7 +40,7 @@ module.exports = function(ServiceInstance) {
       // Save of multiple Services
       ServiceInstance.find(ctx.where, function(err, services) {
         if (err) return next(err);
-        async.each(
+        return async.each(
           services,
           function(instance, callback) {
             serviceManager.onInstanceUpdate(instance, callback);
@@ -56,7 +56,7 @@ module.exports = function(ServiceInstance) {
 
     ctx.Model.find(ctx.where, function(err, instances) {
       if (err) next(err);
-      async.each(
+      return async.each(
         instances,
         function(instance, callback) {
           serviceManager.onInstanceDestroy(instance, callback);
@@ -65,6 +65,24 @@ module.exports = function(ServiceInstance) {
       );
     });
   });
+
+  function recordInstanceInfo(instanceId, instInfo, callback) {
+    ServiceInstance.findById(instanceId, function(err, instance) {
+      if (err) return callback(err);
+      instance.currentDeploymentId = instInfo.commitHash;
+      instance.startTime = new Date();
+      instance.started = true;
+      instance.applicationName = instInfo.appName;
+      instance.npmModules = instInfo.npmModules;
+      instance.PMPort = instInfo.PMPort;
+      instance.containerVersionInfo = instInfo.containerVersionInfo;
+      instance.setSize = instInfo.setSize;
+      instance.agentVersion = instInfo.agentVersion;
+      instance.restartCount = instInfo.restartCount;
+      instance.save(callback);
+    });
+  }
+  ServiceInstance.recordInstanceInfo = recordInstanceInfo;
 
   // Only allow updating ServiceInstance
   ServiceInstance.disableRemoteMethod('create', true);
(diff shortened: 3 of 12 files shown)


server/models/service-instance.js
```diff
@@ -66,6 +67,32 @@ module.exports = function(ServiceInstance) {
...
+      if (err) return callback(err);
+      instance.currentDeploymentId = instInfo.commitHash;
+      instance.startTime = new Date();
+      instance.started = true;
+      instance.applicationName = instInfo.appName;
+      instance.npmModules = instInfo.npmModules;
+      instance.PMPort = instInfo.PMPort;
+      instance.containerVersionInfo = {
+        os: {
+          platform: os.platform(),
```
rmg: Can these runtime details like OS and startTime be passed in as part of `instInfo` instead? I'm picturing a future bug where someone's cluster is left on an older version of Node because all their apps appear to be running in the exact same environment as their cluster controller.

server/models/agent-trace.js
```diff
@@ -0,0 +1,42 @@
+var async = require('async');
+var debug = require('debug')('strong-mesh-models:server:agent-trace');
+var deleteOld = require('./delete-old');
+
+module.exports = function(AgentTrace) {
```
rmg: Can this function be named?

server/models/express-usage-record.js
```diff
@@ -0,0 +1,38 @@
+var debug = require('debug')('strong-mesh-models:server:express-usage');
+var deleteOld = require('./delete-old');
+
+module.exports = function(ExpressUsageRecord) {
```
rmg: This function should be named.

server/models/service-instance.js
kraman: This would be handled different on mesh. But for PM, does this really make a difference?

server/models/express-usage-record.js
```diff
@@ -0,0 +1,38 @@
...
+      serviceInstanceId: instanceId,
+      pid: pid,
+      stopTime: null
+    }};
+
+    ServiceProcess.findOne(q, function(err, proc) {
+      if (err) return callback(err);
+      if (!proc) {
+        debug('Ignoring metrics for unknown worker pid: %d', pid);
+        return deleteOld();
```
rmg: Shouldn't `callback` be passed? Otherwise there appears to be code paths where `recordUsage()` never "returns" (in the async sense).

server/models/service-process.js
```diff
@@ -0,0 +1,132 @@
+var async = require('async');
+var debug = require('debug')('strong-mesh-models:server:service-process');
+
+module.exports = function(ServiceProcess) {
```
rmg: Function should be named.

server/models/express-usage-record.js
kraman: good catch, ill fix that

server/models/service-instance.js
kraman: Good point, I'll make the whole container version info object an option that is passed in

server/models/service-process.js
```diff
@@ -0,0 +1,132 @@
...
+      debug('on exit of %j, save Process: %j', pInfo, err || proc);
+      callback(err);
+    });
+
+    function updateProcess(proc, asyncCb) {
+      if (!proc) {
+        // Ignore process if it is not in the DB. Defensive coding against race
+        // conditions, shouldn't happen, never seen, but lets handle it.
+        debug('exit event for an unknown process: %j', pInfo);
+        return asyncCb();
```
rmg: Doesn't this break `updateChildren()` since it results in its `proc` argument being `undefined`, which it doesn't guard against?

kraman: yep, fixing.

server/models/service-process.js
```diff
@@ -0,0 +1,132 @@
...
+      callback();
+    });
+  }
+  ServiceProcess.recordFork = recordFork;
+
+  function recordExit(instanceId, pInfo, callback) {
+    debug('Process exited: worker id %d pid %d reason %s suicide? %s',
+      pInfo.id, pInfo.pid, pInfo.reason, pInfo.suicide);
+
+    async.waterfall([
```
rmg: More of an "FYI" than a "please change"; one of the conventions I follow is it to explicitly `return` on lines like these so the person reading the code knows that this is the intended end of the function.

server/models/service-process.js
```diff
@@ -0,0 +1,132 @@
...
+  function recordFork(instanceId, pInfo, callback) {
+    debug('Process forked: worker id %d pid %d ppid %d',
+      pInfo.id, pInfo.pid, pInfo.ppid);
+
+    var proc = new ServiceProcess({
+      pid: pInfo.pid,
+      parentPid: pInfo.ppid,
+      workerId: pInfo.id,
+      serviceInstanceId: instanceId,
+      startTime: new Date()
```
rmg: If this API can be used remotely, doesn't that make this `startTime` actually refer to the serialization time (local to the server) rather than the actual start time (local to the actual host running the process)?

kraman: It does, but currently supervisor does not return the actual start time in the event. If its required to report actual time, that should be a separate enhancement req/issue.

test/test-notification-model-updates.js
```diff
@@ -0,0 +1,382 @@
...
+
+    var ServerService = app.models.ServerService;
+    var ServiceInstance = app.models.ServiceInstance;
+    var Executor = app.models.Executor;
+    async.series([
+      ServerService.create.bind(ServerService, s),
+      Executor.create.bind(Executor, e),
+      ServiceInstance.create.bind(ServiceInstance, i)
+    ], function(err) {
+      tt.ok(!err);
```
rmg: Why not use `tt.ifError(err);` instead?

test/test-notification-model-updates.js
```diff
@@ -0,0 +1,382 @@
+var ServiceManager = require('../index').ServiceManager;
+var async = require('async');
+var meshServer = require('../index').meshServer;
+var os = require('os');
+var test = require('tap').test;
+
+test('Test notifications', function(t) {
+  var app = meshServer(new ServiceManager());
```
rmg: This seems like a good argument for making the `serviceManager` argument optional.

kraman: I would like to keep this explicit. The default implementation should only be used during tests. Any real usecase would need to pass in a real implementation.

test/test-notification-model-updates.js
kraman: This wasn't erroring out in one of my tests but I may have just had an old version of tap or assert.
I'll switch it out.

test/test-notification-model-updates.js
```diff
@@ -0,0 +1,382 @@
...
+        addressType: 4
+      }
+    };
+    app.handleModelUpdate(1, notification, function(err) {
+      tt.ok(!err);
+      tt.end();
+    });
+  });
+
+  function checkWorker1(prof, tt) {
```
rmg: Could this test file be split up? It's a little long, and the alternating between function declarations, example data, and actual test code make for a very long and intimidating test suite.

test/test-notification-model-updates.js
```diff
@@ -0,0 +1,382 @@
...
+    });
+  }
+
+  t.test('Notify metrics with current data',
+    notifyMetrics.bind(this, 0));
+  t.test('Notify metrics with stale data',
+    notifyMetrics.bind(this, 10 * 60 * 1000));
+
+  t.test('Check worker 1 metrics', function(tt) {
+    app.models.ServiceMetric.find({}, function(err, metrics) {
```
rmg: Why not `findOne`?

rmg: Nevermind, I see below that is part of the test.

test/test-notification-model-updates.js
```diff
@@ -0,0 +1,382 @@
...
+      isRunning: value
+    };
+    app.handleModelUpdate(1, notification, function(err) {
+      tt.ok(!err);
+      tt.end();
+    });
+  }
+
+  t.test('Notify object-tracking on',
+    notifyProfiling.bind(this, 'object-tracking', true));
```
rmg: I often do this when it only happens once or twice in a file, but with this many calls to `.bind()` would you consider switching to `_.partial` instead?

test/test-notification-model-updates.js
```diff
@@ -0,0 +1,382 @@
...
+  });
+
+  t.test('Check process exit state', function(tt) {
+    var ServiceProcess = app.models.ServiceProcess;
+    ServiceProcess.find({where: {workerId: 0}}, function(err, procs) {
+      tt.ok(!err, 'process records should be found');
+      tt.equal(procs.length, 1, 'only one process should be present');
+      var proc = procs[0];
+      tt.equal(proc.stopReason, 'foo', 'stop reason should match');
+      tt.ok(!!proc.stopTime, 'stop time should be set');
```
rmg: I haven't read line-by-line, but would it be possible to test `rpoc.stopTime` against a range or format instead of just asserting that it is truthy?

test/test-notification-model-updates.js
kraman: Tests execute pretty fast but if you think it would help with readability or something, I can switch...

rmg:
Smattering of comments. Back to you, @kraman.

test/test-notification-model-updates.js
kraman: Certainly can be split. I was debating if I should but had decided to keep it together to see that the entire lifecycle of a process works as expected.

server/models/service-process.js
rmg: Deferring to a separate issue sounds reasonable.

test/test-notification-model-updates.js
rmg: Interesting reversal on test-oriented DI.

test/test-notification-model-updates.js
rmg: `tt.ok(!err);` works fine here, it just doesn't know to report details about `Error` instances when the assertion fails.

test/test-notification-model-updates.js
rmg: Understandable. I'm probably just flinching because of the pain caused by strong-pm's test/test-pmctl-local.js ;-)

test/test-notification-model-updates.js
rmg: Readability was my main concern in this case, yes.

kraman: Hm, looking at API for `_.partial`, The line would look pretty much the same...

```notifyProfiling.bind(this, 'object-tracking', true));```
vs
```_.partial(notifyProfiling, 'object-tracking', true);```

leaving it as-is for now

rmg: Could import it as `var partial = require('lodash').partial;` to make it shorter if you like.

For me the readability improvement comes from seeing right away that it is a partial application rather than some sort of instance method that returns a function.

But I don't care enough to make you change it, that's just my reasoning.

kraman:
@rmg Addressed your comments. Ready for final review.

rmg:
LGTM

kraman:
Rebasing, updating commit messages and squashing fixup! commits

PR #21, opened 2015-03-03
Add Service methods for setting/unsetting environment variables

rmg:
Adds PUT and DELETE handlers for /Services/:id/env/:name to set/unset individual environment variables so that API callers don't need to fetch/modify/put the entire environment object just to make a single change.

Based on #19, so most of the commits are from there.

Connected to strongloop/strong-pm#48

### DIFF
--- a/server/models/server-service.js
+++ b/server/models/server-service.js
@@ -115,4 +115,20 @@ module.exports = function(ServerService) {
     ServerService.app.serviceManager.getDeployment(this, ctx.req, ctx.res);
   }
   ServerService.prototype.getPack = getPack;
+
+  function setEnv(name, value, callback) {
+    this.env = this.env || {};
+    this.env[name] = value;
+    debug('setEnv(%j, %j)', name, value);
+    this.save(callback);
+  }
+  ServerService.prototype.setEnv = setEnv;
+
+  function unsetEnv(name, callback) {
+    this.env = this.env || {};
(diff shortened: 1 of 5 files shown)


sam-github:
Looks OK to me, but reassigning to @kraman because he knows LB better than I.

One comment: you don't implement GET, but I think arc will need it, so that it can see what is currently there and display them, and/or decide to change them.

rmg:
@sam-github I thought about it, but decided that GET would be redundant with just looking at the `.env` member of the Server instance that is probably being used to make the set/unset calls on.

sam-github:
I see, I missed that.

rmg:
Rebased on to master now that #19 has been merged. Should be much easier to review now!

server/models/server-service.js
```diff
@@ -115,4 +115,20 @@ module.exports = function(ServerService) {
     ServerService.app.serviceManager.getDeployment(this, ctx.req, ctx.res);
   }
   ServerService.prototype.getPack = getPack;
+
+  ServerService.prototype.setEnv =
```
kraman: nit, can you switch the way the function is added to prototype to match above methods please

kraman:
Small nit, but LGTM

server/models/server-service.js
rmg: Eww.. :-P

Looks like I also missed the style change (expression to statement) when I rebased.

PR #26, opened 2015-03-06
Fix recording of express metric timestamp

kraman:
Record coming from strong-express-metric module use `timestamp` instead
of `timeStamp`.


rmg:
Could/should we make it accept both forms?

kraman:
@rmg See https://github.com/strongloop/strong-express-metrics/blob/master/index.js#L56
Its hardcoded to be `timestamp`

rmg:
@kraman and up until a couple days ago, it was hardcoded in this repo too, but that didn't stick ;-)

LGTM, with or without accepting both forms.

PR #27, opened 2015-03-09
make service setEnv/unsetEnv return environment

rmg:
The setEnv and unsetEnv endpoints were poorly defined and I think this what they should be.


sam-github:
LGTM

PR #28, opened 2015-03-10
Extract client and CLI from strong-pm

kraman:
Connect to strongloop-internal/scrum-nodeops#311


kraman:
@slnode test please

kraman:
@slnode test please

kraman:
@slnode test please

bin/sl-meshctl.js
```diff
@@ -0,0 +1,486 @@
...
+  ':v(version)',
+  'h(help)',
+  'C:(control)'
+].join(''), argv);
+
+var apiUrl = process.env.STRONGLOOP_MESH_API ||
+  process.env.STRONGLOOP_PM ||
+  exists('pmctl') ||
+  exists(path.join(home, '.strong-pm', 'pmctl')) ||
+  '/var/lib/strong-pm/pmctl' ||
```
sam-github: typo: missing `exists()`, the next line will never be hit as written

bin/sl-meshctl.js
```diff
@@ -0,0 +1,486 @@
...
+var optind = parser.optind();
+
+if (optind < argv.length) {
+  command = argv[optind++];
+}
+
+debug('API Url: %s', apiUrl);
+
+if (!command) {
+  console.error('No command, try `%s --help`.', $0);
```
sam-github: You don't like `status` as default command? Or haven't implemented it yet?

bin/sl-meshctl.txt
```diff
@@ -0,0 +1 @@
+xxx
```
sam-github: didn't get to this yet? should be able to just copy down `sl-pmctl.txt` from strong-pm

client/client.js
```diff
@@ -3,23 +3,53 @@ var assert = require('assert');
...
 var url = require('url');
+var urlDefaults = require('strong-url-defaults');
 var util = require('util');
+var path = require('path');
+
+/**
+ * Mesh API Client
+ *
+ * @param {string} apiUrl URL for Mesh server. Formats supported include:
+ *   * http+unix:///full/path/to/socket
```
sam-github: I'd like this to support partial-socket paths, with no protocol scheme, as well. Right now some of that normalization is in callers of the client.

client/client.js
```diff
@@ -3,23 +3,53 @@ var assert = require('assert');
...
   var endpoint = url.parse(apiUrl);
-  debug('normalize endpoint %j', endpoint);
-
-  if (endpoint.protocol === 'http+unix:') {
+  if (!endpoint.protocol) {
+    this.apiUrl = util.format(
+      'http://unix:%s:%s',
+      path.resolve(apiUrl), apiBasePath
+    );
+  } else if (endpoint.protocol === 'http+unix:') {
```
sam-github: Why invent your own scheme, and not use the one from the request module? Since neither are going to by typed directly into the CLI, and yours just gets rewritten to the request module scheme, seems redundant.

client/client.js
```diff
@@ -35,6 +66,11 @@ function Client(apiUrl) {
   // Populates the cache of models so you can access them with
   // client.models.ModelName.
   client.models();
+
+  // Override and implement the remote methods that use HTTP context because
+  // LB will not provide the correct client implementation.
```
sam-github: I don't understand the above comment.

client/client.js
```diff
@@ -69,6 +105,26 @@ function serviceDestroy(name, callback) {
...
+ *
+ * @param {string|number} serviceNameOrId Service ID or Name.
+ * @param {function} callback Callback function.
+ */
+function serviceFind(serviceNameOrId, callback) {
+  var Service = this.models.ServerService;
+  var filter = {
+    where: {
+      or: [
+        {name: serviceNameOrId},
```
sam-github: You might need a name validator that forbids naming services `1234`

package.json
```diff
@@ -17,14 +17,14 @@
   },
   "main": "index.js",
   "scripts": {
-    "eslint": "eslint client common server test",
-    "jscs": "jscs client common server test",
+    "lint": "eslint client common server test && jscs client common server test",
```
sam-github: To make this work with the coming soon CI support, you should rewrite as `eslint . && jscs .`, then add a .eslintignore and an "excludeFiles" property in your .jscsrc. If your .eslintignore is identical to your .gitignore, then you can do `eslint --ignore-path=.gitignore` (CI will do that, too).

test/test-deploy-endpoints.js
```diff
@@ -33,36 +34,62 @@ test('Create and destroy a service', function(t) {
...
+    });
+  });
 
-      req.write('some data');
+  t.test('deploy a service', function(tt) {
+    var req = service.deploy('application/some-type', function(err, res) {
+      tt.ifError(err);
+      var expResponseBody = 'ok. ServiceID: ' + service.id;
+      res.setEncoding('utf8');
+      res.on('error', t.ifError.bind(t));
```
sam-github: @rmg What do you think, as a node-tap maintainer, of making the methods on `t` be functions, pre-bound to the correct object? I pass `t.end` and `t.ifError` pretty frequently, and its annoying they have to be bound.

test/test-meshctl-ls.json
```diff
@@ -0,0 +1,1388 @@
+{
```
sam-github: are you comparing against this? because its likely to change as any deps change. might be better to assert just on some fields known to be stable.

sam-github: I mean: "as deps of deps change upstream"

test/test-meshctl-ls.render-full
```diff
@@ -0,0 +1,25 @@
+test@0.0.0
```
sam-github: this as well may fail as deps-of-deps have patch releases

bin/sl-meshctl.js
sam-github: status is implemented, it should be default command, I think, unless you've a reason to avoid it?

sam-github:
Looks like its tons of work, but going great. I only skimmed the tests, but it looks like nicely broken down tap tests, with good code reuse in terms of helpers.

bin/sl-meshctl.js
kraman: fixed

bin/sl-meshctl.js
kraman: I had it as default ... https://github.com/strongloop/strong-mesh-models/pull/28/files#diff-3a7a58b347e437190ee1cab5905e5b20R45

bin/sl-meshctl.txt
kraman: yes, almost direct copy. working on this now

client/client.js
kraman: I do support this. Forgot to document. Will update.

client/client.js
kraman: The unix socket scheme used by request is not easy to parse and validate. Using a custom scheme of http+unix:// makes that a lot simpler. http+unix:// is also similar to the http+ssh:// syntax that us used in CLI. If we switch to accepting the request style schema then I will just need to move this code into the CLI which seems unnecessary.

client/client.js
kraman: Methods like `ServerService.getPack` are remote methods defined using the http ctx object. See https://github.com/strongloop/strong-mesh-models/blob/feature/refactor-client/common/models/service.js#L59-59. This means that when you boot up the models, LB tried to fill in the implementation but does it incorrectly. The `client/models/server-service.js` file implements this correctly.

The next problem is load order. `client/models/server-service.js` needs to be loaded after models are booted and attached to the remote datasource, otherwise they will be over-ridden by the incorrect implementation.

I was trying to explain the above in a concise manner.

client/client.js
kraman: Good point. I'll keep that in mind when we start implementing multi-service PM and Mesh

bin/sl-meshctl.js
kraman: done

package.json
kraman: done

test/test-meshctl-ls.json
kraman: Im using `test-meshctl-ls.json`,  `test-meshctl-ls.render-full` and `test-meshctl-ls.render-2` and sample files for testing purposes. The actual dependencies don't matter.

kraman:
@slnode test please

bin/sl-meshctl.js
sam-github: I see. I think this code isn't reachable: command starts with a value, it can only change by getting a new value. You might be able to hit this block by doing `meshctl 0` (a command of `0`, falsy), but in that case the msg is weird, and would be better handled down below, by the `unknown()` command handler.

test/test-deploy-endpoints.js
rmg: Hm.. as convenient as it is, binding is a one-way street and I'd be worried that people are depending on being able to do the binding themselves. It would also interfere with some extension patterns.

How about a tap wrapper that does [lodash.bindAll()](https://lodash.com/docs#bindAll) on `t` before passing it to your function? Or even just `t = _.bindAll(t);` on the first line is pretty unobtrusive.

client/client.js
sam-github: ok

client/client.js
sam-github: Its a subtle point, hard to understand concisely, maybe insert the above comment into the head of one of the implementation files?

test/test-deploy-endpoints.js
sam-github: The `bindAll()` trick is a good one, I didn't know about that lodash function, I'll remember it.

sam-github:
LGTM

sam-github:
Uh... other than my comment about code I think is unreachable, LGTM... I don't need to see this again, though, looks good to go.

kraman:
strongloop/strong-mesh-models/blob/master/common/models/service.js#L19

kraman:
@slnode test please

PR #30, opened 2015-03-17
Browserify support in Client

kraman:


server/server.js
```diff
@@ -1,12 +1,41 @@
...
   var app = module.exports = loopback();
   app.serviceManager = serviceManager;
+  options = options || {};
+
+  for (var k in options) {
+    if (!options.hasOwnProperty(k)) continue;
+    app.set(k, options[k]);
+  }
+
+  assert(!app.get('trace.db.inMemory') && !app.get('trace.db.path'),
```
rmg: Is this right? Crash if either or both are set?

test/test-trace-client-api.js
```diff
@@ -0,0 +1,124 @@
...
+
+  function MockMinkelite() {
+  }
+
+  function shutdown() {
+  }
+  MockMinkelite.prototype.shutdown = shutdown;
+
+  function getMetaTransactions(act, host, pid, callback) {
+    assert.equal(act, 'test-app');
```
rmg: If this mock is used for assertions then these calls should be tracked so that you can assert that they were called.

server/server.js
kraman: Crash if `trace.db.inMemory === false` so we need to persist the DB, but the `trace.db.path` was not provided.

test/test-trace-client-api.js
kraman: I am verifying that the value returned by each of those prototype functions in tests, so I make sure that the correct prototype is run.

server/server.js
rmg: Ya, I see it now.. I think it's the assert instead of `if (...)` that threw me off.

kraman:
Depends on https://github.com/strongloop/loopback-boot/pull/106

kraman:
Merged into tracing PR

PR #34, opened 2015-03-24
meshctl: report pm versions in status

sam-github:


bin/sl-meshctl.js
```diff
@@ -144,6 +144,10 @@ function cmdStatus(instance) {
     fmt(1, 'port', '%s', rsp.port);
     fmt(1, 'base', '%s', rsp.base);
 
+    if (rsp.version) {
```
rmg: Should this check be more elaborate? The format of `rsp.version` theoretically depends on the value in `rsp.version.api`...

rmg: Or `rsp.version.pm`, I suppose.. since this isn't part of mesh models, _per se_, unless it provides a default value.

rmg:
Other the recursive philosophical meta nit.. LGTM.

bin/sl-meshctl.js
sam-github: yeah, depends on the "host" app version. I reworked this so it is a bit more generic, and can work with strong-pm or strong-scheduler (or whatever we call it).

There's kindof a fundamental problem, here, which is the client-side display code, above, is tied to the server data structure. It would be nice if whatever is returned from the server was just dumped... but json dumps look awful. I guess the server side could return both the json data... as well as some kind of template that the client could expand when it wanted to display it... 

Ouph, way too much work for now.

Anyhow, PTAL at the these, see if its better now.

rmg: I think a JSON dump would be perfectly fine in this case.. it's more of a debug/diagnostics command anyway..

bin/sl-meshctl.js
```diff
@@ -144,6 +144,12 @@ function cmdStatus(instance) {
...
+      fmt(1, 'version');
+      fmt(2, rsp.version.name, 'v%s', rsp.version.version);
```
rmg: Now these object paths just look weird.. maybe version should just be a flat object and the UI can just dump it as key: value...

```js
rsp.version = {
  api: require('strong-mesh-models/package.json').version,
  'strong-pm': require('../package.json').version,
};
```

Leave the `v` prefix out of the printing at let the server put it in, that way it is free to put free-form strings in place of these values if desired.
```js
fmt(1, 'version');
Object.keys(rsp.version).forEach(function(k) {
  fmt(2, '%s: %s', k, rsp.version[k]);
});
```

rmg:
Sorry, this turned into a bit of a bike shed :-/

bin/sl-meshctl.js
sam-github: free-form strings can't be interacted with, I'd rather leave formatting to the client. I'll rework as key/value.

PR #36, opened 2015-03-28
runner refactor

sam-github:
connect to strongloop-internal/scrum-nodeops#307

### DIFF
--- a/server/config.json
+++ b/server/config.json
@@ -13,11 +13,11 @@
     },
     "json": {
       "strict": false,
-      "limit": "200mb"
+      "limit": "200kb"
     },
     "urlencoded": {
       "extended": true,
-      "limit": "200mb"
+      "limit": "200kb"
     },
     "cors": {
       "origin": true,
--- a/server/config.json
+++ b/server/config.json
@@ -13,11 +13,11 @@
     },
     "json": {
       "strict": false,
-      "limit": "100kb"
+      "limit": "200mb"
     },
     "urlencoded": {
       "extended": true,
-      "limit": "100kb"
+      "limit": "200mb"
     },
     "cors": {
       "origin": true,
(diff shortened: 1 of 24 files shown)


server/server.js
```diff
@@ -103,6 +103,7 @@ function server(serviceManager, options) {
     // take care to not callback until any actions are complete.
     if (!callback) {
       callback = function(err) {
+        // XXX(sam) shouldn't err of an err-first already be an Error() object?
         if (err) throw Error(err);
```
sam-github: @kraman I think the `Error: Error: ...` we saw in meshctl demo might be from here, or code like this.

server/server.js
```diff
@@ -157,6 +158,10 @@ function server(serviceManager, options) {
       case 'express:usage-record':
         ExpressUsageRecord.recordUsage(instanceId, uInfo, callback);
         break;
+      case 'status:wd':
```
sam-github: @kraman should I do this, or let it pass through as an unrecognized request?

server/server.js
kraman: Yes, thats a bug but not the one we were seeing in my demo. A throw here would cause process to exit.

server/server.js
kraman: Not using this notification at all. The data that this provides already comes with other notifications like fork, listening etc.

sam-github: I don't think the current working directory is known, and I don't think the PWD value is known if we don't do an explicit env-get command.

sam-github:
@kraman tests pass, please review

server/server.js
sam-github: I left the switch in so the command wouldn't get reported as Unknown.

server/config.json
```diff
@@ -13,11 +13,11 @@
     },
     "json": {
       "strict": false,
-      "limit": "100kb"
+      "limit": "200mb"
```
kraman: do we still need to increase this limit? Having a large limit can lead to a JSON based DOS attack on the API

sam-github:
Merged onto https://github.com/strongloop/strong-mesh-models/pull/38

PR #38, opened 2015-03-31
Update CLI to support multiple services

kraman:
Connect to strongloop-internal/scrum-nodeops#315


bin/sl-meshctl.js
```diff
@@ -123,213 +125,332 @@ function unknown() {
...
   var vars = mandatorySome('K=V');
   var env = _.reduce(vars, extractKeyValue, {});
-  instance.envSet(env, function(err, response) {
+
+  client.serviceFind(targetService, function(err, service) {
     dieIf(err);
-    console.log('Environment updated: %j', response.message);
+    service.setEnvs(env, function(err, responses) {
+      dieIf(err);
+      // Fail if any of the instances failed to update
```
rmg: This comment suggests that this code is aware of the multiple instances, which suggests this is where we would need to implement rollback logic. Shouldn't that be on the server?

kraman: The client makes no assumption of rollback or not. It just prints any errors that were sent down from the server.

rmg: What I meant was the instance level errors should probably be rolled into a single error message, otherwise the client's knowledge of the instances suggests that it might be responsible for resolving them later.

bin/sl-meshctl.js
```diff
@@ -123,213 +125,332 @@ function unknown() {
...
+      for (var i in responses) {
+        if (!responses.hasOwnProperty(i)) continue;
+        dieIf(responses[i].error);
```
rmg: It looks like only the first error will be displayed before dying, is that correct?

kraman: Good catch. I'll fix that.

bin/sl-meshctl.js
kraman: Hm, yes. I would prefer to leave this in here and add an XXX for when we add rollback support on server. The current way its implemented just delegates the update to each instance.

bin/sl-meshctl.txt
```diff
@@ -29,25 +29,24 @@ be overridden to use an existing private key instead of an agent by setting the
 `SSH_KEY` environment variable to the path of the private key to be used.
 
 Commands:
-  status                  Report status, the default command.
+  status [SRV_ID]         Report status for Service SRV_ID, or all services if
+        no SRV_ID is provided. This is the default command.
```
rmg: Indentation is wrong for this short of a description.

common/models/api.js
```diff
@@ -0,0 +1,12 @@
+module.exports = function Api(Api) {
+  Api.remoteMethod(
+    'shutdown',
+    {
+      http: {path: '/shutdown', verb: 'post'},
+      isStatic: true,
+      accepts: [],
+      returns: {arg: 'response', type: 'object', root: true},
+      description: 'Shutdown this executor.'
```
rmg: This description suggests this method should be on the Executor model.

server/models/service-instance.js
```diff
@@ -84,6 +84,158 @@ module.exports = function extendServiceInstance(ServiceInstance) {
...
+    });
+  }
+  ServiceInstance.prototype.runCommand = runCommand;
+
+  function _appCommand(obj, callback) {
+    obj.sub = obj.cmd;
+    obj.cmd = 'current';
+    this.runCommand(obj, callback);
+  }
+  ServiceInstance.prototype._appCommand = _appCommand;
```
rmg: This appears to be used extensively by ServiceProcess internals, perhaps the `_` prefix should be removed?

server/models/service-process.js
```diff
@@ -205,4 +205,120 @@ module.exports = function extendServiceProcess(ServiceProcess) {
...
   ServiceProcess.prototype._getActAndHost = _getActAndHost;
+
+  /**
+   * Start tracking objects on a worker.
+   *
+   * @param {function} callback Callback function.
+   */
+  function startObjectTracking(callback) {
+    var self = this;
+    self.serviceInstance(function(err, instance) {
```
rmg: It might be worth surfacing a class/static variant of `ServiceInstance#_appCommand` to reduce the boilerplate in these 6 very similar methods.

rmg:
I see a lot of room for cleanup and refactoring, but I think it is mostly intentional so that the refactoring is more "pure". Other than the early exit on set/unset-env failure, LGTM.

Oh, and needs a rebase, in case you didn't notice the merge conflict :-)

bin/sl-meshctl.txt
kraman: @rmg Can you elaborate? this is a 2 line description ...

common/models/api.js
kraman: Changed to `Shutdown this server`. Intention is to shutdown whatever service is providing this API

server/models/service-instance.js
kraman: yah, makes sense

bin/sl-meshctl.txt
kraman: @rmg leaving this for now since it matches indentation in the rest of the file.

rmg:
LGTM. :shipit:

sam-github:
@kraman check out https://github.com/strongloop/strong-mesh-models/commit/73ce7f8a280614790aec1c33f38a808d757b9445 and the XXX commits.

sam-github:
@kraman Check out e5b836c, if you don't have other things on your mind! :-) I could have sworn the branch tests passed before I started rebasing, but now, even when going back to the original branch, I still find failures, so I think I must have got confused and run the pm tests by mistake. ouch.

I'll manually verify cpu and heap profiling, so be confident skipping that test is ok for the moment. For express usage record, I guess that won't matter until #41 is merged.

I'll probably merge once tests are green on CI.

sam-github:
test please

sam-github:
@rmg, I fixed this: http://ci.strongloop.com/job/strong-mesh-models.pr/182/label=amazon-64-0.10/tapResults/, but it keeps coming up in the tests. Argh.

sam-github:
everything passes for me on OS X with a fresh checkout (except ssh2, it seems to be  having some auth failures, not sure why).

sam-github:
test please

rmg:
@slnode test please

rmg:
@sam-github CI just built clean.. maybe Jenkins just likes me more than you :-P

PR #42, opened 2015-04-30
server: add ServerService.serviceFindOrCreate

sam-github:
These methods were already available from Client, but the pm needs to
call them as well.

connected to strongloop-internal/scrum-nodeops#511

### DIFF
--- a/server/models/server-service.js
+++ b/server/models/server-service.js
@@ -70,6 +70,50 @@ module.exports = function extendServerService(ServerService) {
   }
   ServerService.setServiceCommit = setServiceCommit;
 
+  // XXX(sam) serviceFindOrCreate, serviceCreate, serviceFind were almost
+  // verbatim ripped from client/client.js: there should be a way of sharing the
+  // implementation, perhaps this code can be attached to ServerService in
+  // common/models and be available identically on client and server as a js
+  // method?
+  function serviceFindOrCreate(name, scale, callback) {
+    var self = this;
+    self.serviceFind(name, function(err, service) {
+      if (err && err.statusCode !== 404) return callback(err);
+      if (service) return callback(null, service);
+      self.serviceCreate(name, scale, callback);
+    });
+  }
+  ServerService.serviceFindOrCreate = serviceFindOrCreate;
+
+  function serviceCreate(name, scale, callback) {
+    var Service = this; /* eslint consistent-this:0 */
+
+    if (scale == null) scale = 1;
+    var svc = {name: name, _groups: [{id: 1, name: 'default', scale: scale}]};
+    Service.create(svc, callback);
+
+    return this;
+  }
+  ServerService.serviceCreate = serviceCreate;
+
+  function serviceFind(serviceNameOrId, callback) {
+    var Service = this; /* eslint consistent-this:0 */
+
+    var filter = {order: ['id ASC']};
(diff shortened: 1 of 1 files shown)


client/models/server-service.js
```diff
@@ -73,7 +73,7 @@ module.exports = function extServerService(ServerService) {
         var instance = instances[i];
         if (!instance.containerVersionInfo)
           continue;
-        var container = instance.containerVersionInfo.container;
+        var container = instance.containerVersionInfo.container; //
```
rmg: typo comment?

common/models/api.json
```diff
@@ -32,7 +32,8 @@
     "driverStatus": {
       "type": "string",
       "required": true,
-      "description": "The hostname of this executor"
+      "XXX": "value is always 'started', what is it for?",
```
rmg: I think there's a convention of `_comment` as the key for things like this.

server/models/server-service.js
```diff
@@ -70,6 +70,52 @@ module.exports = function extendServerService(ServerService) {
...
+    self.serviceFind(name, function(err, service) {
+      if (err && err.statusCode !== 404) return callback(err);
+      if (service) return callback(null, service);
+      self.serviceCreate(name, scale, callback);
+    });
+  }
+  ServerService.serviceFindOrCreate = serviceFindOrCreate;
+
+  function serviceCreate(name, scale, callback) {
+    var self = this;
```
rmg: You alias `this` twice.. `Service` makes sense as a "commenting variable", but `self` is just redundant.

server/models/server-service.js
```diff
@@ -70,6 +70,52 @@ module.exports = function extendServerService(ServerService) {
...
+    if (scale == null) scale = 1;
+    var svc = {name: name, _groups: [{id: 1, name: 'default', scale: scale}]};
+    Service.create(svc, callback);
+
+    return self;
+  }
+  ServerService.serviceCreate = serviceCreate;
+
+  function serviceFind(serviceNameOrId, callback) {
+    var self = this;
```
rmg: Ditto here regarding aliasing of `this`.

rmg:
I half expected based on the PR description to see these methods added to common/models and removed from the client...

Actually, in both cases I'm not sure why they are needed at all.. it isn't much of an abstraction over:
```js
var filter = {where: {name: name}};
var service = {name: name, _groups: [{id: 1, name: 'default', scale: scale}]};
Service.findOrCreate(filter, service, callback);
```

rmg:
Other than the comment and `this` aliasing nits, LGTM to merge.

common/models/api.json
sam-github: I committed a bunch of temp changes, I'm sorry.

server/models/server-service.js
sam-github: keeps lint happy, but I'll add an anti-lint comment and simplify the code. lint hates `Service = this`, it expects all this to be assigned to self...

sam-github:
re: https://github.com/strongloop/strong-mesh-models/pull/42#issuecomment-97894086 I think that might work, but then it would change the client API, and I don't want to break that.

`Service.findOrCreate` <--- no such method exists, that's what this PR introduces

rmg:
Is `Service` not a `PersistedModel`?

PR #43, opened 2015-04-30
list listening ports in status report

rmg:
Connect to strongloop-internal/scrum-nodeops#507

### DIFF
--- a/bin/sl-meshctl.js
+++ b/bin/sl-meshctl.js
@@ -204,11 +204,12 @@ function printServiceStatus(service) {
       ]);
     }
     console.log('Instances:\n%s', table(instanceTable, {
-      align: ['c', 'c', 'c', 'c', 'c']
+      align: ['c', 'c', 'c', 'c', 'c', 'c']
     }));
 
     var processTable = [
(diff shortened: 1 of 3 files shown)


bin/sl-meshctl.js
```diff
@@ -223,6 +224,7 @@ function printServiceStatus(service) {
         proc.displayId,
         proc.pid,
         proc.workerId,
+        _.pluck(proc.listeningSockets, 'port').join(','),
```
sam-github: Will this properly ignore unix domain and UDP sockets? unix domain won't have a port, not sure if pluck will return undefined or ignore that one, but UDP will have a port, I  think.

sam-github:
basically LGTM, though I'm not sure its robust against non-TCP ports. because the node docs are so terrible, a manual test would probably be in order

rmg:
I wasn't going to worry about the domain socket case because it seems even more unlikely than listening on multiple ports for our theoretical user.. but I've fixed it now. PTAL.

sam-github:
LGTM

PR #45, opened 2015-05-01
server: persist instance.cpus to trigger resize

sam-github:
Only the set-size command was actually changing a running instance's
size. Updates to the size were being notified as onInstanceUpdate to pm,
but it wouldn't apply them to the running instance (doing so requires
sending a set-size command).

Now, if you don't want persistence, a set-size command is sent, which is
non-persistent.

If you do want persistence, the model is updated, that update goes to
pm, which then updates both the instance sl-run CLI options AND sends a
set-size command to a running instance (if size has changed). Because of
this, the persistent branch that sends set-size is no longer necessary.

Also, some effort was being made to capture non-numeric values for cpus,
and turn them into some default value. This code is confusing and
unused, its now gone. The size can be set to whatever is wanted (whether
that size has any effect on strong-cluster-control is a different
matter).

Probably, validation should occur, but that doesn't happen right now.

connected to strongloop-internal/scrum-nodeops#537

### DIFF
--- a/common/models/service-instance.json
+++ b/common/models/service-instance.json
@@ -53,7 +53,6 @@
     },
     "cpus": {
       "type": "string",
-      "default": "cpus",
       "description": "The number of cpus to use when starting the service on this instance"
     },
     "applicationName": {
--- a/bin/sl-meshctl.js
+++ b/bin/sl-meshctl.js
@@ -387,12 +387,13 @@ function cmdRollingRestart(client) {
 function cmdSetClusterSize(client) {
   var targetService = mandatory('service');
   var size = mandatory('size');
+  var persist = false;
 
   client.serviceFind(targetService, function(err, service) {
     dieIf(err);
     service.setClusterSize(
       size,
-      false,
+      persist,
       printResponse.bind(null, service, null)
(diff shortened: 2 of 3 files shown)


common/models/service-instance.json
```diff
@@ -53,7 +53,6 @@
     },
     "cpus": {
       "type": "string",
-      "default": "cpus",
```
rmg: Why remove the default?

rmg:
The change looks pretty straight forward. LGTM.

common/models/service-instance.json
sam-github: Because this default gets persisted in the DB as the intended size... overriding any more dynamic notion of what the default size should be. Like `STRONGLOOP_CLUSTER`. Now, if you set a per-instance cpus value, it will have it, if you don't, it will be undefined, and pm decides what size to use.

rmg:
The `if (foo) return; return;` structure threw me off at first (misread and didn't see the implicit `else`), but that could have just been from viewing it in as a diff instead of the final code.

LGTM. :shipit:

PR #47, opened 2015-05-02
meshctl: clarify usage message

sam-github:
Identify more clearly the scope of commands, and don't use ID...
services, executors, processes and workers all have IDs, its too
confusing.

connected to strongloop-internal/scrum-nodeops#508

### DIFF
--- a/bin/sl-meshctl.txt
+++ b/bin/sl-meshctl.txt
@@ -28,62 +28,29 @@ overridden by setting an `SSH_USER` environment variable. The authentication can
 be overridden to use an existing private key instead of an agent by setting the
 `SSH_KEY` environment variable to the path of the private key to be used.
 
-Commands:
-  status [SVC]            Report status for service SVC, or all services if
-        no SVC is provided. This is the default command.
 
-  create NAME [SCALE]     Create service named NAME, SCALE defaults to 1.
-  ls                      List services.
-  remove SVC              Remove a service SVC.
+Global commands: apply to the process manager itself
 
+  info                    Information about the process manager.
   shutdown                Shutdown the process manager.
 
-  start SVC               Start service SVC.
-  stop SVC                Hard stop service SVC.
-  soft-stop SVC           Soft stop service SVC.
-  restart SVC             Hard stop and restart service SVC with new config.
-  soft-restart SVC        Soft stop and restart service SVC with new config.
-        "Soft" stops notify workers they are being disconnected, and give them
-        a grace period for any existing connections to finish. "Hard" stops
-        kill the supervisor and its workers with `SIGTERM`.
+
+Service commands: apply to a specific service
+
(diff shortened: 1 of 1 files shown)


crandmck:
Separating the sub-commands by functional area is an improvement.  Within each of those three sections (Global commands, Service commands, and Worker commands) I suggest you order the commands alphabetically.  I realize that you've ordered them in what seems like a functional order, but that's not obvious and in some cases, there is no clear order (e.g. why is `objects-start` listed before `cpu-start/stop`).  In fact, beyond the first few commands the order is essentially random, or at least not clear.  

This is a command reference (not task-oriented user doc) and the subcommands should be in alphabetical order.  If you want to have a short summary of the most-used commands (for service commands, for example-- start, stop, create, status) that would be OK.  

Finally, I'm still not convinced that calling the identifier "service" adds anything.  As far as I can tell, it's really just an identifier that enables PM to work with a specific deployed instance of an app.  It's only used by PM (and with the corresponding `slc` commands).  The  application becomes available over the network at a port number determined by PM unrelated to the "service" identifier.  Calling it "service"  makes me think it's something more than that.  I think "app-name", "deployment identifier", or something along those lines might be better.  But at this point, it might be too late and too much work to change it.  Anyway, I'll deal with the fallout and implications.

rmg:
@crandmck I think the terminology is an open discussion, so it might make sense to sit on this PR a little bit longer until we have clearer direction on that. BTW, the ordering is most likely the order they were implemented ;-)

sam-github:
This PR doesn't change the name, we can do elsewhere.

I agree on alphabetical, the reference is long enough that makes the most sense, thanks.

I'm OK with going to App instead of service, but it may confuse things. Its a bit like calling a box with a blender inside it a "blender". That totally makes sense in some situations when you don't care about the   distinction, like when buying a blender at a store. But in some situations, it makes less sense.

A service is a box into which apps can be placed, and when they are placed there, they then can have actions applied: start, stop, remove all (an empty box), list the multiple versions of an app that are associated with that service. roll back some or all instances to an earlier version of the app. In the latter case, the service remains the same, but the app inside it changes. So maybe the service and the app aren't the same thing?

An app could be described as the tarball, docker image, etc. The deployable entity. In which case, calling the thing that is running and managing the app "app" is confusing - thus the blender vs box holding a blender analogy. But, maybe that's not really that confusing... it will be clear in the context, and using "app" definitely reduces the amount of terminology we introduce.

All of which is to say, I'm not strongly one way, or the other. Its easy to replace all mention of SVC/service with APP/app in the help message. I just need to be told to do that.

@crandmck convince Al, Issac, or Chanda (or all three). I'm OK either way, and am not the one to make the decision!

crandmck:
@sam-github Thanks.  I can see both (several?) sides of the argument... 

In any case, I can't make the decision, either.  I'll try to get the above folks in the discussion and make a decision.

PR #48, opened 2015-05-04
fix null listening addresses in meshctl status

rmg:
Node 0.12 and iojs, when no bind address is specified, give null as
the listening address instead of '0.0.0.0', so as a quick fix we'll
just do that substitution ourselves.

Connect to strongloop-internal/scrum-nodeops#542

### DIFF
--- a/bin/sl-meshctl.js
+++ b/bin/sl-meshctl.js
@@ -7,13 +7,13 @@ var _ = require('lodash');
 var assert = require('assert');
 var concat = require('concat-stream');
 var debug = require('debug')('strong-mesh-client:meshctl');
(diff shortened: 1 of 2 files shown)


sam-github:
LGTM

sam-github:
I wonder if the change from 0.0.0.0 to null has something to do with better IPv6 support, so that the default can be  `0.0.0.0` or the IPv6 equivalent.

rmg:
The 0.10 behaviour was to return `::` or `0.0.0.0` depending on which stack was used. Note that the `family` field is still there, and still shows either `IPv4` or `IPv6`.

PR #49, opened 2015-05-04
meshctl: fix message splitting

sam-github:
`slc ctl -C http://no-such.com` was triggering the message splitting in
the middle of the host name, so split only on '. ', which is what
loopback uses for long errors.


rmg:
LGTM.

PR #50, opened 2015-05-04
meshctl: prompt on unsupported REST calls

sam-github:
When single-app strongloop hits multiapp, print a more specific message
than 'Error: unsupported'.

connected to strongloop-internal/scrum-nodeops#543


rmg:
LGTM.

PR #52, opened 2015-05-05
Add setEnvs remote method to Service model

rmg:
Back ported from master/6.x

I'd like to release this as 5.1.0 so that strongloop/strong-arc#1226 can be simplified to use the same API it will use when strong-mesh-client is upgraded to use strong-mesh-models@6.

/cc @jtary


sam-github:
LGTM

PR #54, opened 2015-05-06
meshctl: ls should be listed as global

sam-github:


sam-github:
@crandmck  liked it on chat

PR #56, opened 2015-05-08
watchdog auto-stop support

sam-github:
connected to strongloop-internal/scrum-nodeops#558

### DIFF
--- a/common/models/service-process.json
+++ b/common/models/service-process.json
@@ -44,6 +44,11 @@
       "default": 0,
       "description": "Watchdog timeout, in milliseconds. (0 = watchdog disabled)"
     },
+    "watchdogStallout": {
+      "type": "number",
+      "default": 0,
+      "description": "Watchdog stallout, in stall counts. (0 = stall disabled)"
+    },
     "isSnapshotting": {
       "type": "boolean",
       "required": true,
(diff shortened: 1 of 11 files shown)


sam-github:
@jtary you are implementing arc side, right? We should review to make sure this works for you.

common/models/service-process.json
```diff
@@ -44,6 +44,11 @@
       "default": 0,
       "description": "Watchdog timeout, in milliseconds. (0 = watchdog disabled)"
     },
+    "watchdogStallout": {
```
rmg: Isn't this basically the same as `ProfileData.coun({where: {ServiceProcessId: pid}})` (or something like that).

sam-github: No, its not a count of the number of profiles. When you do `cpu-start PROC TIMEOUT STALLOUT`, the messages goes down to the target, which echoes back a status update event: https://github.com/strongloop/strong-supervisor/blob/master/lib/targetctl.js#L103-L110

It is those isProfiling, timeout, stallout values that then go into this model.

Basically, this is so when arc lists a process, it knows if its profiling, and if it is profiling, what mode its profiling in (normal, timeout, timeout+stallout).

If you can think of some better wording to make this clear, I can update the description, but I'm fighting the data format here. Also, note line 42-46 just above: the description is identical, `/timeout/stallout/g`.

rmg:
LGTM.

PR #57, opened 2015-05-16
Properly revive processes on fork

rmg:
When upserting a ServiceProcess instance for a `fork` event, clear the `stopReason` and `stopTime` so that they are properly marked as running if they were previously marked as dead.

Also some small tweaks to some tests to work with an upgrade of tap.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   "scripts": {
     "lint": "eslint . && jscs .",
     "pretest": "eslint . && jscs .",
-    "test": "tap --tap --stdout --stderr --timeout=300 test/test-*.js",
+    "test": "tap --bail --timeout=300 test/test-*.js",
     "cover": "test/cover.sh"
   },
   "dependencies": {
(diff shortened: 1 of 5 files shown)


rmg:
@sam-github this is required for strongloop/strong-pm#189

sam-github:
LGTM

but one thing that's not clear... when pm reconnects to a docker instance, it is going to get fork events for *all* living workers, right?

And also, the master process? I'm worried the children will be marked as alive, but the parent will still be marked 'stopped'.

sam-github:
Also, all these tests are red, but they are all [green on master](http://ci.strongloop.com/job/strong-mesh-models/).

rmg:
> but one thing that's not clear... when pm reconnects to a docker instance, it is going to get fork events for all living workers, right?

Yes, the docker driver will emit fake fork events for all living workers when it connects..

> And also, the master process? I'm worried the children will be marked as alive, but the parent will still be marked 'stopped'.

This is done by the `started` message already, but without this PR it results in creating a new "master" process each time instead of resurrecting the previous one when appropriate.

The code before this PR assumed that any process being "forked" is new and therefore has the default "null" values for stopReason/stopTime.

Basically this PR is making the findOrCreate-on-fork behaviour work as intended.

rmg:
@sam-github had to expand the changes a bit, can you take another look before I merge?

sam-github:
LGTM

PR #59, opened 2015-05-20
Add support to turn on/off tracing for an instance

kraman:
Connect to strongloop-internal/scrum-nodeops#459

### DIFF
--- a/common/models/service-instance.js
+++ b/common/models/service-instance.js
@@ -112,6 +112,26 @@ module.exports = function(ServiceInstance) {
   );
 
   ServiceInstance.remoteMethod(
+    'tracingStart',
+    {
+      http: {path: '/tracingStart', verb: 'post'},
+      isStatic: false,
+      returns: {arg: 'response', type: 'object', root: true},
+      description: 'Restart all workers with tracing on.'
+    }
+  );
+
+  ServiceInstance.remoteMethod(
+    'tracingStop',
+    {
+      http: {path: '/tracingStop', verb: 'post'},
+      isStatic: false,
+      returns: {arg: 'response', type: 'object', root: true},
+      description: 'Restart all workers with tracing off.'
+    }
+  );
+
+  ServiceInstance.remoteMethod(
     'npmModuleList',
     {
       http: {path: '/npmModuleList', verb: 'post'},
--- a/common/models/service-instance.json
+++ b/common/models/service-instance.json
@@ -55,6 +55,11 @@
       "type": "string",
       "description": "The number of cpus to use when starting the service on this instance"
     },
+    "tracingEnabled": {
+      "type": "boolean",
+      "default": false,
+      "description": "Indicates if tracing is enabled on this instance"
+    },
     "applicationName": {
       "type": "string",
       "description": "The application name that strong-agent uses to report its metrics"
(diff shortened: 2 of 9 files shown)


common/models/service-instance.js
```diff
@@ -112,6 +112,26 @@ module.exports = function(ServiceInstance) {
   );
 
   ServiceInstance.remoteMethod(
+    'tracingStart',
+    {
+      http: {path: '/tracingStart', verb: 'post'},
+      isStatic: false,
+      returns: {arg: 'response', type: 'object', root: true},
+      description: 'Start tracing on all workers.'
```
sam-github: I would say "Restart all workers with tracing on", and ditto for below.

common/models/service-instance.json
```diff
@@ -55,6 +55,11 @@
       "type": "string",
       "description": "The number of cpus to use when starting the service on this instance"
     },
+    "enableTracing": {
```
sam-github: I'm OK with this, but it sounds like a verb/action not a boolean state... would tracingEnabled be better?

server/models/service-instance.js
```diff
@@ -7,9 +7,15 @@ module.exports = function extendServiceInstance(ServiceInstance) {
     function(ctx, _, next) {
       debug('updateAttributes with %j', ctx.args.data);
       // When updating the instance via REST, only allow changes to cpus
-      ctx.args.data = {
-        cpus: ctx.args.data.cpus
-      };
+      // and enableTracing
+      var reqData = ctx.args.data;
+      ctx.args.data = {};
+      if (reqData.cpus)
```
sam-github: Are the conditionals needed?

server/models/service-process.js
```diff
@@ -128,6 +128,23 @@ module.exports = function extendServiceProcess(ServiceProcess) {
   }
   ServiceProcess.recordProfilingState = recordProfilingState;
 
+  function recordStatusWdUpdate(instanceId, pInfo, callback) {
+    function updateProcessStatus(proc, asyncCb) {
+      proc.isTracing = pInfo.isTracing || false;
```
sam-github: `proc.isTracing = !!pInfo.isTracing` would make more sense to me if you want to coerce to a boolean, but I'd hope that loopback would coerce, since field type in model is boolean, in which case a simple assignment is sufficient.

server/models/service-process.js
```diff
@@ -128,6 +128,23 @@ module.exports = function extendServiceProcess(ServiceProcess) {
...
+      proc.save(asyncCb);
+    }
+
+    return async.waterfall([
+      _findProcess(instanceId, +pInfo.id, pInfo.pid, +pInfo.pst),
```
sam-github: I find function definition before the async hard to read in comparison to presenting code in the order it is executed.

server/models/service-process.js
```diff
@@ -128,6 +128,23 @@ module.exports = function extendServiceProcess(ServiceProcess) {
...
+
+      updateProcessStatus
+    ], function(err, proc) {
+      debug('Process entry updated: %j', proc);
```
sam-github: `err || proc`?

sam-github: so errors get printed

sam-github:
A couple minor comments, other than that, LGTM

server/models/service-instance.js
kraman: Yah, I need to set the fields only if they are not null otherwise the update might fail

common/models/service-instance.json
kraman: ok

common/models/service-instance.js
kraman: ok

server/models/service-process.js
kraman: Not sure if LB coerces the value being stored

PR #63, opened 2015-05-25
trace-object: records are stringified now

bnoordhuis:
strong-supervisor now emits the trace object record as a string, as a
performance optimization.

Refs: https://github.com/strongloop/strong-supervisor/pull/129

R=@sam-github


sam-github:
LGTM

PR #64, opened 2015-05-26
Fix status and tracing commands

kraman:
Connect to strongloop-internal/scrum-nodeops#627

### DIFF
--- a/bin/sl-meshctl.txt
+++ b/bin/sl-meshctl.txt
@@ -84,6 +84,10 @@ services using `ls` or `status`.
         a grace period for any existing connections to finish. "Hard" stops
         kill the supervisor and its workers with `SIGTERM`.
 
+  tracing-start SVC       Restart all workers with tracing on.
+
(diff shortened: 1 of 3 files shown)


bin/sl-meshctl.txt
```diff
@@ -84,6 +84,10 @@ services using `ls` or `status`.
         a grace period for any existing connections to finish. "Hard" stops
         kill the supervisor and its workers with `SIGTERM`.
 
+  tracing-start SVC       Restart all workers with tracing on.
```
sam-github: Eventually, this should be `SVC | SVC.EXEC`, but this is fine for strong-pm, where its the same.

sam-github:
LGTM

bin/sl-meshctl.txt
kraman: yah, that was my intention. altough it accepts SVC.EXEC I didnt want to complicate the help info

PR #66, opened 2015-05-28
Model updates for strong-central

kraman:
Connect to strongloop-internal/scrum-nodeops#421

### DIFF
--- a/client/client.js
+++ b/client/client.js
@@ -405,3 +405,20 @@ function expressMetricsEndpointDetail(modelOrUri, windowStartTime, callback) {
   );
 }
 Client.prototype.expressMetricsEndpointDetail = expressMetricsEndpointDetail;
+
+function executorCreate(driver, callback) {
+  this.models.Executor.create({
+    driver: driver || 'executor'
+  }, callback);
+}
+Client.prototype.executorCreate = executorCreate;
+
+function executorList(callback) {
+  this.models.Executor.find({}, callback);
+}
+Client.prototype.executorList = executorList;
+
+function executorDestroy(id, callback) {
+  this.models.Executor.deleteById(id, callback);
+}
+Client.prototype.executorDestroy = executorDestroy;
--- a/server/auth.js
+++ b/server/auth.js
@@ -1,0 +1,64 @@
+var auth = require('http-auth');
+var crypto = require('crypto');
+var fmt = require('util').format;
+
+module.exports = makeAuthMiddleware;
+module.exports.parse = parseAuth;
+
+var authMethods = {
+  basic: makeBasicAuth,
+  digest: makeDigestAuth,
+  none: makeNoAuth,
+};
+
+function makeAuthMiddleware(pmAuthStr) {
+  var parts = parseAuth(pmAuthStr);
+  return authMethods[parts.scheme](parts.user, parts.realm, parts.pass);
+}
+
+var AUTH_REGEXP = /^(basic|digest):(.+):(.+)$/;
+function parseAuth(pmAuthStr) {
+  var parts = AUTH_REGEXP.exec(pmAuthStr) ||
+              AUTH_REGEXP.exec('basic:' + pmAuthStr);
+  var result = {
+    scheme: parts ? parts[1] : 'none',
+    user: parts && parts[2],
+    pass: parts && parts[3],
+    realm: 'strong-pm',
+  };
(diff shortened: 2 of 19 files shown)


kraman:
Non-backward compat changes will be on a different PR

sam-github:
I think sl-meshadm should be installed: https://github.com/strongloop/strong-mesh-models/blob/master/package.json#L21  (will make things easier for us, and be invisible to our customers).

I could with for a little less cut-n-paste between the two bin/sl-*.js utilities: https://github.com/strongloop/strong-mesh-models/blob/feature/strong-central/bin/sl-meshadm.js#L69-L89 and https://github.com/strongloop/strong-mesh-models/blob/feature/strong-central/bin/sl-meshadm.js#L149-L186 look like they could be extracted into `lib/cli-util.js`, but we could do that later.

server/auth.js
```diff
@@ -0,0 +1,64 @@
...
+}
+
+function makeNoAuth() {
+  return function noop(req, res, next) {
+    next();
+  };
+}
+
+function makeBasicAuth(user, realm, pass) {
+  console.log('PM: enabling HTTP authentication (Basic)');
```
rmg: log message should be updated, it's not PM specific anymore

server/auth.js
```diff
@@ -0,0 +1,64 @@
...
+  var basicAuth = auth.basic({realm: realm}, checkCredentials);
+  return auth.connect(basicAuth);
+
+  function checkCredentials(maybeUser, maybePassword, cb) {
+    cb(maybeUser === user && maybePassword === pass);
+  }
+}
+
+function makeDigestAuth(user, realm, pass) {
+  console.log('PM: enabling HTTP authentication (Digest)');
```
rmg: log message should be updated, it's not PM specific anymore

rmg:
The hooks could be made more backwards compatible with a `manager.fn.length === 2` check to see how many arguments the handler is expecting.. A little annoying, but could be very handy for releasing these sorts of improvements out earlier rather than later.

kraman:
@rmg is fn.length available in strict mode?

rmg:
@kraman it sure is. Readonly property of the Function prototype since ECMAScript 1. Safer than semicolons, I'd say ;-)

kraman:
@rmg PTAL

client/client.js
```diff
@@ -405,3 +405,20 @@ function expressMetricsEndpointDetail(modelOrUri, windowStartTime, callback) {
   );
 }
 Client.prototype.expressMetricsEndpointDetail = expressMetricsEndpointDetail;
+
+function executorCreate(driver, callback) {
```
rmg: This is almost certainly going to be expanded, perhaps it should start off as an object so it doesn't need to be changed later?

server/models/executor.js
```diff
@@ -0,0 +1,30 @@
+var async = require('async');
+
+module.exports = function(Executor) {
+  Executor.observe('after save', function(ctx, next) {
+    var serviceManager = Executor.app.serviceManager;
+    if (ctx.instance) {
+      if (serviceManager.onExecutorUpdate.length === 2) {
```
rmg: Not necessary here, since this handler is being defined for the first time in this PR.

rmg:
For the hooks, is the `ctx.isNewInstance` different than the hooks just calling `instance.isNewRecord()`? I guess that might not work for `after save`, but I would expect it to work for `before save`..

rmg:
Couple minor things to cleanup, and a general question, otherwise LGTM.

server/models/executor.js
kraman: Just staying consistent with the other hooks

kraman:
@rmg yes `ctx.isNewInstance` is different and the thing that should be used in hooks

client/client.js
kraman: @rmg I am not sure what parameters will go in here. I would prefer to evolve this function as we expand on functionality in mesh. Im sure it will look different by the time its time for out first release

rmg: If the first parameter is an object instead of a string, then the general prototype of the method won't have to change if you add parameters.

server/models/executor.js
rmg: I don't think it is worth it if the other checks are scheduled to be removed when the major version bumps.

rmg:
I'm ok if this gets merged as is, it's not the last time this code will be modified :-)

PR #67, opened 2015-05-28
Direct profile start as action

sam-github:
connected to strongloop-internal/scrum-nodeops#635


kraman:
Other than the query, LGTM

sam-github:
@kraman that query is a verbatim copy of https://github.com/strongloop/strong-mesh-models/blob/master/client/client.js#L313-L316, so if its broken here, its broken there. I'd like to leave it (or I can add a commit that adds the and to both?).

kraman:
ok, lets leave it for now

PR #65, opened 2015-05-27
meshctl: assert instance is well-formed

sam-github:
Attempt to find source of this difficult to reproduce bug:

strong-mesh-models/node_modules/text-table/index.js:59
    return m ? m.index + 1 : c.length;
                              ^
TypeError: Cannot read property 'length' of null
    at dotindex (strong-mesh-models/node_modules/text-table/index.js:59:31)
    at strong-mesh-models/node_modules/text-table/index.js:11:21
    at Array.forEach (native)
    at forEach (strong-mesh-models/node_modules/text-table/index.js:73:31)
    at strong-mesh-models/node_modules/text-table/index.js:10:9
    at Array.reduce (native)
    at reduce (strong-mesh-models/node_modules/text-table/index.js:63:30)
    at module.exports (strong-mesh-models/node_modules/text-table/index.js:9:20)
    at strong-mesh-models/bin/sl-meshctl.js:214:35
    at strong-mesh-models/client/models/server-service.js:100:9

PR #68, opened 2015-05-28
fix over aggressive assert

rmg:
Asserts added in 94ef1e4 in #65 broke pmctl when accessing an instance
with cluster size of 0, which is the initial state for apps that are
dockerized by strong-pm.

This was causing strong-pm's pmctl tests fail for me.


sam-github:
LGTM

PR #69, opened 2015-05-29
report driver metadata for each instance

rmg:
Connect to strongloop-internal/scrum-nodeops#485


test/test-meshctl-status.js
```diff
@@ -36,8 +36,8 @@ test('Test status command', function(t) {
     'Environment variables:',
     '  No environment variables defined',
     'Instances:',
-    '    Version  Agent version  Cluster size',
-    '     1.2.3       7.8.9            3',
+    '    Version  Agent version  Cluster size  Driver metadata',
+    '     1.2.3       7.8.9            3              {}',
```
sam-github: `N/A` would make more sense than `{}`, I think.

What does it look like for docker?

rmg: It will print `N/A` if there is no metadata, the mock instance used for the tests just happens to have an empty object for its metadata.

rmg: I'm working on a PR for PM that will result in the docker driver printing something like `{container: '12ab34cd56ef'}`

kraman:
Metadata may contain non-user-visible stuff as well. how do you plan to filter those out?

rmg:
@kraman at the moment, it contains nothing.. I'm not sure if/how we want to make the distinction between user visible and not. Maybe make it look for `driverMeta.user`?

kraman:
@rmg yah, that works.

rmg:
@slnode test please

rmg:
@sam-github @kraman LGTY now?

rmg:
Output from the docker driver looks like this:
```
Service ID: 1
Service Name: one
Environment variables:
  No environment variables defined
Instances:
    Version  Agent version  Cluster size         Driver metadata
     4.1.1       1.5.1            8       {"container":"59356153be26"}
Processes:
      ID    PID  WID  Listening Ports  Tracking objects?  CPU profiling?  Tracing?
     1.1.1   1    0
     1.1.6   6    1     0.0.0.0:3001
     1.1.8   8    2     0.0.0.0:3001
    1.1.10   10   3     0.0.0.0:3001
    1.1.12   12   4     0.0.0.0:3001
    1.1.14   14   5     0.0.0.0:3001
    1.1.16   16   6     0.0.0.0:3001
    1.1.18   18   7     0.0.0.0:3001
    1.1.20   20   8     0.0.0.0:3001
```

sam-github:
the output isn't so pretty, but I can see the genericity of it is worthwhile. anyhow, LGTM

PR #70, opened 2015-05-29
fix race condition in process update processing

rmg:
If multiple updates to the same process were happening concurrently
then the last handler to call `proc.save()` would overwrite the
changes made by the other handler. This is because `proc.save()` saves
the entire object instead of only updating the changed fields.

This manifested in listenening sockets being dropped randomly, as
reported in strongloop-internal/scrum-nodeops#639, because the
listening and status:wd events are sent at almost the same time and
are sometimes processed concurrently.

To reduce the problems caused by this, `proc.updateAttributes()` is
used wherever possible so that concurrent saves to different attributes
are at least safe.

Note that there is still a problem if multiple updates to the same
attributes happen concurrently. Fortunately, that does not tend to
happen in practice.

Connect to strongloop-internal/scrum-nodeops#639

### DIFF
--- a/server/models/service-process.js
+++ b/server/models/service-process.js
@@ -39,9 +39,10 @@ module.exports = function extendServiceProcess(ServiceProcess) {
         return asyncCb();
       }
       if (proc.stopTime == null) {
-        proc.stopTime = new Date();
-        proc.stopReason = pInfo.reason;
-        return proc.save(asyncCb);
+        return proc.updateAttributes({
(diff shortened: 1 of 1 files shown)


rmg:
I couldn't think of how to cover this in the tests, but the problem went from reproducible on every attempt with direct or docker driver to unreproducible after repeated attempts.

sam-github:
LGTM

PR #74, opened 2015-06-05
Use only HTTP for control, not local sockets

sam-github:
connected to strongloop-internal/scrum-nodeops#444


rmg:
Removes a dependency and a bunch of `fs.existsSync` calls, I don't even care what it does, LGTM!

PR #75, opened 2015-06-07
Update models to match supervisor notification updates

kraman:
Connect to strongloop-internal/scrum-nodeops#673


sam-github:
LGTM

I like the wid... every time I see `id` I think ... what kind? loopback model id? something else? This is much clearer.

PR #77, opened 2015-06-12
Model updates for strong-central

kraman:
connect to strongloop-internal/scrum-nodeops#675

### DIFF
--- a/common/models/service.js
+++ b/common/models/service.js
@@ -72,14 +72,6 @@ module.exports = function(Service) {
       description: 'Unset environment variables'
     });
 
-    this.remoteMethod('getPack', {
-      isStatic: false,
-      http: [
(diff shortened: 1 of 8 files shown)


rmg:
The removal of `getPack` seems incomplete (didn't remove references from tests), but I'm not sure if it is a useful change or not. Won't we need something like those endpoints once central is storing some sort of "release" blob and it gets pushed/pulled around to different executors?

kraman:
fixing tests

kraman:
Executor will not be using the REST API exposed by mesh-models, in Central there will be a `/artifacts` endpoint to retrieve this file.

rmg:
@slnode test please

rmg:
LGTM.

PR #87, opened 2015-06-30
meshctl: preserve = in env var values

### LINKED ISSUE #86: env-set does not preserve values if they have = in them
Originally reported by @jrschumacher on strongloop/chat gitter:
```
slc pmctl env-set app DB_URL="mongodb://user:pass@db.example.com:27017/app?replicaSet=12345"
Service "app" environment updated
    Name      Value
    NODE_ENV  production
    DB_URL    mongodb://user:pass@db.example.com:27017/app?replicaSet
```

rmg:
URLs, such as those used by database connectors, often have = in them
and they should be preserved.

Closes #86

### DIFF
--- a/bin/sl-meshctl.js
+++ b/bin/sl-meshctl.js
@@ -597,7 +597,7 @@ function cmdEnvSet(client) {
         pair, $0);
       process.exit(1);
     }
-    store[kv[0]] = kv[1];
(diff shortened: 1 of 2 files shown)


rmg:
@kraman @sam-github FYI I plan on merging this and as soon as CI passes and then back porting to 7.x and making a patch release.

PR #88, opened 2015-07-02
Add option to mesh-models to use external db

kraman:
Connect to strongloop-internal/scrum-nodeops#409


server/server.js
```diff
@@ -28,20 +28,28 @@ function server(serviceManager, minkelite, options) {
...
   boot(app, __dirname);
 
   app.start = function(callback) {
-    // start the web server
-    app._server = app.listen(function() {
-      var addr = this.address();
-      app.emit('started', addr.port);
-      console.log('Web server listening at port: %s', addr.port);
-      if (callback) return callback(null, addr.port);
+    app.dataSources.db.automigrate(function(err) {
```
rmg: Just to confirm, this _is_ a no-op and not an error like "memory store doesn't support migration" when an external store isn't provided, right?

rmg:
This PR appears to completely remove the ability to control the DB via the `STRONGLOOP_MESH_DB` environment variable, which appeared to already provide the functionality this PR is trying to add.

Couldn't this PR be reduced to just adding the few lines to run automigrate() without breaking backwards compatibility?

kraman:
@rmg, using STRONGLOOP_MESH_DB would only work if the database connector was in the package dependencies of mesh-models. The way I have it now, you can create the datasource outside mesh-models and pass it in. That way we dont pollute mesh-models with connectors which it doesnt need.

server/server.js
kraman: Yes, this is a no-op on memory datasource

rmg:
Hm... On the one hand, it clearly breaks compatibility... but on the other hand, the previous interface was also very flawed in that even if the parent module added a dependency on the connector it would break if the dependencies were flattened or hoisted up a level.

kraman:
@rmg It does break compatibility but nothing can use the old interface in any case.

rmg:
Ya, I'm just thinking through the implications aloud..

server/server.js
kraman: @raymondfeng Should this be `autoupdate` or is `automigrate` correctly used here?
We dont want data to be wiped every time ...

test/test-service-external-db.js
```diff
@@ -0,0 +1,33 @@
+var meshServer = require('../index').meshServer;
+var ServiceManager = require('../index').ServiceManager;
+var SQLite3 = require('loopback-connector-sqlite3');
+var test = require('tap').test;
+
+test('Service environment variable manipulation', function(t) {
+  var manager = new ServiceManager();
+  var DataSource = require('loopback-datasource-juggler').DataSource;
+  var db = new DataSource(SQLite3);
```
rmg: Is there any chance of using a mock or spy here instead of a real connector? The tests down below don't actually check that _this_ datasource is the one that was used.

rmg:
The tests have room for improvement, but the rest of my concerns aren't really addressable because they're just necessary changes.

LGTM once the test is extended.

test/test-service-external-db.js
kraman: @rmg Hm, good point about not testing if datasource is correct. I will add a test for that but mocking the whole connector is not practical.

rmg: I'm not sure how much it actually does.. I was hoping a handful of methods that could just return true or something.. or even if they just error out, like make `automigrate()` fail and then check for the error?

kraman: Most of what happens in mesh-models is LB booting the app and attaching models and initializing them.

kraman:
@rmg added a test to make sure the correct Datasource is attached to the model.
Tests are failing because loopback-connector-sqlite3 seems to be stale on CI. PTAL

server/server.js
raymondfeng: If you want to keep date, autoupdate should be used.

raymondfeng: Without the models argument, all models will be migrated. You can be explicit by calling:
```js
ds.autoupdate(['Model1', 'Model2'], function(err) {
});
```

rmg:
@slnode test please

rmg:
@slnode test please

kraman:
@slnode test please

rmg:
@kraman :shipit:

kraman:
@rmg added a minor update to allow specification of db filepath for memory db. PTAL

kraman:
@slnode test please

rmg:
@slnode test please

rmg:
@kraman LGTM, :shipit: 
http://ci.strongloop.com/job/strong-mesh-models.pr/319/

PR #89, opened 2015-07-02
Fix process lookup for recording smart profile data

kraman:
Connect to strongloop-internal/scrum-nodeops#753


rmg:
LGTM.

Is there a similarly trivial change that can be made to the tests so that they catch this?

kraman:
@rmg no tests seem to have written for this. Will need to add a new one for smart-profiling at some point

PR #91, opened 2015-07-09
Inject db configuration instead of datasource

kraman:
Connect to strongloop-internal/scrum-nodeops#409

### DIFF
--- a/test/test-service-external-db.js
+++ b/test/test-service-external-db.js
@@ -1,34 +1,42 @@
 var meshServer = require('../index').meshServer;
 var ServiceManager = require('../index').ServiceManager;
 var SQLite3 = require('loopback-connector-sqlite3');
+var mktmpdir = require('mktmpdir');
 var test = require('tap').test;
+var path = require('path');
 
 test('Service environment variable manipulation', function(t) {
-  var manager = new ServiceManager();
-  var DataSource = require('loopback-datasource-juggler').DataSource;
-  var db = new DataSource(SQLite3);
+  mktmpdir(function(err, dir, done) {
+    t.ifError(err);
+    if (err) return done();
+    t.on('end', done);
 
-  var server = meshServer(manager, null, {
-    db: db
-  });
-  server.set('port', 0);
+    var manager = new ServiceManager();
+    var server = meshServer(manager, null, {
+      db: {
+        connector: SQLite3,
+        file: path.join(dir, 'test.db'),
+      }
+    });
+    server.set('port', 0);
 
-  var Service = null;
+    var Service = null;
 
-  t.test('server setup', function(tt) {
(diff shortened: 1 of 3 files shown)


test/test-service-external-db.js
```diff
@@ -19,7 +18,7 @@ test('Service environment variable manipulation', function(t) {
     server.start(function onStart(err) {
       tt.ifError(err, 'server started successfully');
       Service = server.models.ServerService;
-      tt.equal(Service.dataSource, db);
+      tt.equal(Service.dataSource.connector.name, 'sqlite3');
```
rmg: Can we add an extra assertion here that the _default_ dataSource is not `'sqlite3'` so that if the default every changes this test won't become invalid (and hide bugs)?

rmg:
It's a pretty subtle change.. only one small concern, so LGTM once it is addressed.

test/test-service-external-db.js
kraman: @rmg Not sure what you mean by default dataSource. AFAIK Loopback doesnt have a concept of a default datasource... it has to be explicitly specified in the model-config.json.

kraman:
Waiting for https://github.com/strongloop/loopback-connector-sqlite3/pull/5

kraman:
Also added backward compat which should take care of failing PM builds

rmg:
@slnode test please

test/test-service-external-db.js
rmg: strong-mesh-models will use the memory connector if one isn't given. My concern here is that in the future if the default changes to sqlite3 and the feature for allowing an override breaks, this test will still pass and we won't notice the override is broken.

rmg:
My nit about test paranoia isn't that important. LGTM, once CI comes back green.

rmg:
@kraman seems CI isn't happy about sqlite3.. looks like that missing table again..

kraman:
@slnode test please

kraman:
@slnode test please

test/test-service-external-db.js
```diff
@@ -1,34 +1,41 @@
...
+var mktmpdir = require('mktmpdir');
 var test = require('tap').test;
+var path = require('path');
 
 test('Service environment variable manipulation', function(t) {
-  var manager = new ServiceManager();
-  var DataSource = require('loopback-datasource-juggler').DataSource;
-  var db = new DataSource(SQLite3);
+  mktmpdir(function(err, dir, done) {
+    t.ifError(err);
```
rmg: Should add this so that if mktmpdir fails it doesn't pollute $CWD
```js
if (err) {
  return done();
}
```

rmg:
One small suggestion, otherwise, LGTM.

PR #93, opened 2015-07-13
Use updateAttributes instead of saving whole model

kraman:
Connect to strongloop-internal/scrum-nodeops#677

### DIFF
--- a/server/models/service-instance.js
+++ b/server/models/service-instance.js
@@ -122,14 +122,15 @@ module.exports = function extendServiceInstance(ServiceInstance) {
         return callback(Error(
           fmt('Instance with ID %s not found', instanceId)));
 
+      var updates = {};
       if (instInfo.master && instInfo.master.setSize) {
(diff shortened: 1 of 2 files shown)


rmg:
The whitespace in the last hunk of the diff looks a little weird, but otherwise LGTM.

sam-github:
LGTM

server/models/service-instance.js
```diff
@@ -140,9 +141,9 @@ module.exports = function extendServiceInstance(ServiceInstance) {
       if (!instance)
         return callback(Error(fmt('Instance with ID %s not found', instanceId
           )));
-
-      instance.applicationName = instInfo.appName;
-      return instance.save(callback);
+      return instance.updateAttributes(
+        {applicationName: instInfo.appName}, callback
```
rmg: Now that I'm not on my phone.. this is the line split that looked odd to me. But it's just whitespace, so I can always fix it myself later if I really want to :-)

PR #94, opened 2015-07-14
Undoc SCALE option for service create

kraman:
Connect to strongloop-internal/scrum-nodeops#776


rmg:
Commit message doesn't say why this is being done... too confusing to users?

Otherwise, LGTM.

PR #96, opened 2015-07-17
meshctl: report server name in info output

sam-github:
Allows strong-pm and strong-central to be distinguished by the info
command.

connected to strongloop-internal/scrum-nodeops#805


kraman:
LGTM

PR #97, opened 2015-07-17
Add executor destroy hooks

kraman:
Connect to strongloop-internal/scrum-nodeops#800


sam-github:
does package.json need a minor version bump?

LGTM

kraman:
It should not, the default service-manager.js which the PM and Central inherit from contains a no-op method which will be run if not customized.

PR #98, opened 2015-07-17
Bug 808

kraman:
Connect to strongloop-internal/scrum-nodeops#808


sam-github:
This is because in central, a deployment may never have been given to a driver - unlike pm, where a deployment is immediately handed off to a driver?

LGTM

kraman:
@sam-github its because this information is only available after a supervisor been started for that instance atleast once.

PR #101, opened 2015-08-01
meshctl: order status list by service ID

sam-github:
printServiceStatus() was being called in order of service ID, but since
it makes an async call to a REST API, the actual order of services
depended on the order in which the HTTP calls came back, which is
somewhat random.


kraman:
Dont see you ordering the list by Id. Is having `async.eachSeries` enough?

kraman:
Code change itself LGTM but have not verified ordering.

sam-github:
The services are always returned ordered: https://github.com/strongloop/strong-mesh-models/blob/master/client/client.js#L146

PR #102, opened 2015-08-07
Use numeric IDs for gateway

kraman:
Connect to strongloop-internal/scrum-nodeops#674


kraman:
@slnode test please

rmg:
I'm not a fan of numeric IDs, especially how they seem to go viral.. but the infection seems to have already set in, so LGTM I guess.

PR #103, opened 2015-08-21
fix regression in process hooks

rmg:
New hooks were injected into the process update chain in f4cbe6a674ab1e3a96f24ef2b848238cc3ac29aa but the default implementations did not forward their results, breaking things.

### DIFF
--- a/server/models/service-process.js
+++ b/server/models/service-process.js
@@ -46,7 +46,9 @@ module.exports = function extendServiceProcess(ServiceProcess) {
           stopReason: pInfo.reason,
         }, function(err, proc) {
           if (err) return asyncCb(err);
-          serviceManager.onProcessExit(proc, asyncCb);
+          serviceManager.onProcessExit(proc, function(err) {
+            return asyncCb(err, proc);
+          });
         });
       }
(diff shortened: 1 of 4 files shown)


server/service-manager.js
```diff
@@ -60,13 +60,13 @@ ServiceManager.prototype.onGatewayDestroy = onGatewayDestroy;
 
 function onProcessListen(proc, callback) {
   debug('onProcessListen(%j)', proc);
-  setImmediate(callback);
+  setImmediate(callback, null, proc);
```
rmg: Strictly speaking this one isn't necessary since the place it was injected wasn't forwarding the result, but I think this is a safer default hook anyway.

sam-github:
I'm not sure its supposed to forward the arguments, I think the bug might be in the code which assumes it does. commented here: https://github.com/strongloop/strong-mesh-models/commit/f4cbe6a674ab1e3a96f24ef2b848238cc3ac29aa#commitcomment-12835592

rmg:
@slnode test please

kraman:
:+1:

rmg:
@slnode test please

PR #105, opened 2015-09-08
Move process, metric etc. under instance

kraman:
Connect to strongloop-internal/scrum-nodeops#872

* Hide models other than service, gateway, instance and executor
* Add endpoints to access metrics and process etc under instance
* Add stopTime field to Instance
* Mount explorer only on 'started' event


common/models/service-instance.json
```diff
@@ -24,7 +24,12 @@
     },
     "startTime": {
       "type": "date",
-      "description": "The ID of the service group this executor belongs to"
+      "description": "Timestamp when instance was started"
+    },
+    "stopTime": {
+      "type": "date",
+      "description": "Timestamp when instance was stopped (cloud usecases)",
+      "default": "1/1/1970 00:00:00 GMT+0"
```
rmg: Shouldn't this default to `null` if it is going to be doubling as a state flag?

(sorry for the double comment, the first one was on the commit not the PR diff)

server/server.js
```diff
@@ -15,7 +15,7 @@ var url = require('url');
  * @param {object} options Options object
  */
 function server(serviceManager, minkelite, options) {
-  var app = module.exports = loopback();
```
rmg: This must have been a fun bug to track down!

client/models/service-instance.js
```diff
@@ -2,15 +2,6 @@ var request = require('request');
...
 
 module.exports = function(ServiceInstance) {
-  function applyPatch(target, fileData, callback) {
-    return this._appCommand({
-      cmd: 'patch',
-      target: target,
-      patch: fileData
-    }, callback);
-  }
-  ServiceInstance.prototype.applyPatch = applyPatch;
```
rmg: This appears to be part of the JS api and not just the remote API, which would require a semver major bump. Is there a way not to remove this method?

server/models/service-instance.js
```diff
@@ -321,6 +321,23 @@ module.exports = function extendServiceInstance(ServiceInstance) {
   }
   ServiceInstance.prototype.logDump = logDump;
 
+  [
+    'agentTraces', 'expressUsageRecords', 'serviceMetrics',
```
rmg: This would be more maintenance friendly if it were 1 item per line (and with the trailing comma on the last item).

server/models/service-instance.js
```diff
@@ -321,6 +321,23 @@ module.exports = function extendServiceInstance(ServiceInstance) {
...
+    'getMetaTransactions', 'getTransaction', 'getTimeline',
+    'getTrace', 'stopObjectTracking', 'startObjectTracking',
+    'startCpuProfiling', 'stopCpuProfiling', 'heapSnapshot',
+    'queryCapabilities', 'queryCapabilitiesAll', 'applyPatch'
+  ].forEach(function(methodName) {
+    ServiceInstance.prototype[methodName] = function(pk) {
```
rmg: I'm a fan of metaprogramming in general, but I don't think it buys us much in this case since the argument massaging to account for differences ends up eating the difference.

Maybe instead of a loop, write a 1 arg and 2 arg variant of a helper and just call it for each one?

```js
proxy2(ServiceInstance, 'processes', 'getTrace');
proxy1(ServiceInstance, 'processes', 'getTimeline');
// ...
```

test/test-instance-profiling.js
```diff
@@ -55,12 +54,10 @@ test('Check heap and cpu profiling populates Profile models', function(t) {
...
       t.test('heap snapshot delete', function(tt) {
         TestServiceManager.prototype.onCtlRequest = assertCtlRequests(tt, 1, 1);
-        instance.processes({where: {id: 1}}, function(err, processes) {
+        instance.heapSnapshot(1, function(err, res) {
           tt.ifError(err);
 
-          processes[0].heapSnapshot(function(err, res) {
-            tt.ifError(err);
-
+          setImmediate(function() {
```
rmg: Is this `setImmediate()` necessary? If so, why? Is it covering up a bug? Feel free to answer in the form of a comment in the code if appropriate.

common/models/service-instance.js
```diff
@@ -141,4 +141,311 @@ module.exports = function(ServiceInstance) {
...
   );
+
+  ServiceInstance.remoteMethod(
+    'agentTraces',
+    {
+      http: {path: '/processes/:pk/agentTraces', verb: 'get'},
+      isStatic: false,
+      accepts: [
+        {
+          arg: 'pk',
```
rmg: This seems to break convention. Is there a reason this isn't `id` here and in the path?

rmg:
Left some comments. Nothing too earth shattering, just some minor cleanups.

common/models/service-instance.json
kraman: Doing a find query for `null` over the Rest API is very painful. Searching for an expected default is much simpler. Chose `1/1/1970 00:00:00 GMT+0`.

client/models/service-instance.js
kraman: Im changing a lot of the REST API in this PR. A major version bump is probably required.

common/models/service-instance.js
kraman: this is a sub-route of instance i.e. `/instances/:id/processes/:pk/agentTraces` The `:id` key is already being used so I'm using `:pk` instead.

client/models/service-instance.js
kraman: The JS api does not change. The same method is not implemented as a remote method on ServiceInstance. https://github.com/strongloop/strong-mesh-models/blob/feature/heroku/common/models/service-instance.js#L428. So instance.applyPatch(...) will continue to work.

rmg:
@slnode test please

common/models/service-instance.json
rmg: That's unfortunate.. can we compromise on at least using a real [date format](https://xkcd.com/1179/)?

Should be `1970-01-01T00:00:00.000Z`, and I propose that any datastore that chokes on that doesn't meet our minimum requirements.

rmg:
@slnode test please

rmg:
LGTM, hopefully with a sane date format.

A rebase should reset the stale CI statuses and (assuming I've finally solved the problem) result in a green build.

kraman:
@slnode test please

PR #108, opened 2015-09-13
Add dbWatch template hooks to server models

Setogit:
Template dbWatcher hooks added to server models for initial end-to-end testing.  Existing loopback connector based notification is still intact.

### DIFF
--- a/server/boot/extend-models.js
+++ b/server/boot/extend-models.js
@@ -1,25 +1,29 @@
-var async = require('async');
+'use strict';
+
 var agentTrace = require('../models/agent-trace');
 var Api = require('../models/api');
+var async = require('async');
 var executor = require('../models/executor');
 var expressUsageRecord = require('../models/express-usage-record');
 var gateway = require('../models/gateway');
 var profileData = require('../models/profile-data');
-var serverService = require('../models/server-service');
 var serviceInstance = require('../models/service-instance');
 var serviceMetric = require('../models/service-metric');
 var serviceProcess = require('../models/service-process');
+var serverService = require('../models/server-service');
 
 module.exports = function(server) {
   Api(server.models.Api);
   agentTrace(server.models.AgentTrace);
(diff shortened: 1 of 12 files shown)


server/models/agent-trace.js
```diff
@@ -2,7 +2,14 @@ var async = require('async');
 var debug = require('debug')('strong-mesh-models:server:agent-trace');
 var deleteOld = require('./delete-old');
 
-module.exports = function extendAgentTrace(AgentTrace) {
+exports = module.exports = extendAgentTrace;
+exports.modelWatcher = modelWatcher;
+
+function modelWatcher(msg) {
+  console.log('------------------------ agenttrace:', msg);
```
kraman: Left over debug statement?

server/boot/extend-models.js
```diff
@@ -18,4 +20,27 @@ module.exports = function(server) {
...
+    {name: 'executor', instance: executor},
+    {name: 'expressusagerecord', instance: expressUsageRecord},
+    {name: 'profiledata', instance: profileData},
+    {name: 'serviceinstance', instance: serviceInstance},
+    {name: 'serverservice', instance: serverService},
+    {name: 'servicenetric', instance: serviceMetric},
+    {name: 'serviceprocess', instance: serviceProcess},
+    {name: 'gateway', instance: gateway},
+  ];
+  server.on('ready', function(dbWatcher) {
```
kraman: Where does this event come from and at what point is it emitted? The watchers will need to be set up after DB, Driver etc are crated but before the driver and gateway re-connect calls. Perhaps this should be an method on the mesh server that can be called when we need it?

kraman: This should probably also check if dbWatcher is passed in. It would be null for non-postgres cases...right?

Setogit:
Moved the dbWatch initialization to ServiceManager.  serviceDestroy change is still useful.

PR #111, opened 2015-09-17
Use dbWatcher-based notification

Setogit:
connect to strongloop-internal/scrum-nodeops#899

connect to strongloop-internal/scrum-nodeops#949

### DIFF
--- a/client/client.js
+++ b/client/client.js
@@ -174,12 +174,9 @@ function serviceFind(serviceNameOrId, callback) {
 
   var filter = {order: ['id ASC']};
   if (serviceNameOrId) {
-    filter.where = {
-      or: [
-        {name: serviceNameOrId},
-        {id: serviceNameOrId}
-      ]
-    };
+    filter.where = (typeof serviceNameOrId === 'string') ?
+        {name: serviceNameOrId} :
+        {or: [{name: serviceNameOrId}, {id: serviceNameOrId}]};
   }
   Service.findOne(filter, callback);
 }
@@ -187,14 +184,9 @@ Client.prototype.serviceFind = serviceFind;
 
 function groupCreate(serviceNameOrId, groupName, scale, callback) {
   var Service = this.models.ServerService;
-  var filter = {
-    where: {
-      or: [
-        {name: serviceNameOrId},
-        {id: serviceNameOrId}
-      ]
-    }
-  };
+  var filter = (typeof serviceNameOrId === 'string') ?
+      {where: {name: serviceNameOrId}} :
+      {where: {or: [{name: serviceNameOrId}, {id: serviceNameOrId}]}};
   Service.findOne(filter, function(err, service) {
     if (!service) {
       // We could create it, but then it wouldn't have a default. Until further
@@ -227,14 +219,9 @@ Client.prototype.groupCreate = groupCreate;
 function instanceList(serviceNameOrId, callback) {
   var Service = this.models.ServerService;
   var ServiceInstance = this.models.ServiceInstance;
-  var filter = {
-    where: {
-      or: [
-        {name: serviceNameOrId},
-        {id: serviceNameOrId}
-      ]
-    }
-  };
+  var filter = (typeof serviceNameOrId === 'string') ?
+      {where: {name: serviceNameOrId}} :
+      {where: {or: [{name: serviceNameOrId}, {id: serviceNameOrId}]}};
   Service.findOne(filter, function(err, service) {
     if (err || !service) return callback(err, service);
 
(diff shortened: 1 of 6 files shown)


client/client.js
```diff
@@ -150,7 +150,8 @@ Client.prototype.serviceList = serviceList;
 function serviceDestroy(nameOrId, callback) {
   var Service = this.models.ServerService;
 
-  var q = {where: {or: [{name: nameOrId}, {id: nameOrId}]}};
+  var q = (typeof nameOrId === 'string') ? {where: {name: nameOrId}} :
+      {where: {or: [{name: nameOrId}, {id: nameOrId}]}};
```
kraman: Should update all the other places in client.js that use the same filter style. Eg: service, group create/find/delete locations.

server/models/executor.js
```diff
@@ -1,14 +1,53 @@
...
+
+    if (serviceManager._dbWatcher) {
+      if (!shouldWatch(serviceManager, name)) {
+        setImmediate(next);
+        return;
+      }
+      var watcherCtx = {
+        'modelName': name,
+        'watcher': serviceManager._dbWatcher,
+        'saveFun': saveObserver,
```
kraman: minor nit, I usually use `saveFn` instead of `saveFun` :)

server/models/executor.js
```diff
@@ -1,14 +1,53 @@
...
+        'saveNext': next,
+        'deleteFun': deleteObserver,
+        'modelInst': serviceManager._meshApp.models.Executor,
+      };
+      instModelWatcher(watcherCtx);
+    }
+    saveObserver(ctx, next);
```
kraman: won't this cause double invocation of the saveObserver once the model watcher is registered?

server/models/executor.js
```diff
@@ -1,14 +1,53 @@
...
+    if (serviceManager._dbWatcher) {
+      if (!shouldWatch(serviceManager, name)) {
+        setImmediate(next);
+        return;
+      }
+      var watcherCtx = {
+        'modelName': name,
+        'watcher': serviceManager._dbWatcher,
+        'saveFun': saveObserver,
+        'saveNext': next,
```
kraman: What are you using `saveNext` for?

server/models/executor.js
```diff
@@ -1,14 +1,53 @@
...
+        'deleteFun': deleteObserver,
+        'modelInst': serviceManager._meshApp.models.Executor,
```
kraman: This is the same as just using `Executor` defined on line 7 above. Why refer to serviceManager?

server/models/executor.js
```diff
@@ -1,14 +1,53 @@
...
   // vs data.
 
+  var name = 'executor';
+
   Executor.observe('after save', function(ctx, next) {
     var serviceManager = Executor.app.serviceManager;
-    var instance = ctx.instance || ctx.currentInstance;
+    debug('------------------ Executor SAVE %j.', ctx.where);
+
+    if (serviceManager._dbWatcher) {
```
kraman: Is `serviceManager._dbWatcher` were instead accessed from `Executor.app.get(...)` then you could probably do this logic outside the definition of `Executor.observe` and make the code simpler.

server/models/executor.js
```diff
@@ -1,14 +1,53 @@
...
+var debug = require('debug')('strong-mesh-models:service-manager');
+var instModelWatcher = require('../../lib/helper').instModelWatcher;
+var shouldWatch = require('../../lib/helper').shouldWatch;
 
 module.exports = function(Executor) {
 
   // Refer to server-service.js for description of instance vs currentInstance
   // vs data.
 
+  var name = 'executor';
```
kraman: Why not just get this value from `Executor`? Im sure loopback has something like `Executor.modelName` (need to check docs).

client/client.js
Setogit: What's the "place" ?  <Model>.findOne call?

server/models/executor.js
Setogit: No.  Once watcher is registered, shouldWatch returns false.

server/models/executor.js
Setogit: This is still experimental.  In this version, both saveObserser part of the original Executor.observe callback *and* next are done in the dbWatcher event listener.  This version of implementation is experimental because the saveNext is set once and re-used every time which is bogus.

server/models/executor.js
Setogit: changed

server/models/executor.js
Setogit: right.  this was a left-over from the last edit.

server/models/executor.js
Setogit: replaced.

server/models/executor.js
Setogit: Good idea, yes, that'd be one of the cleanups to be done once the logic is proven to work.

server/models/executor.js
Setogit: `Executor.modelName` is it.  replaced.

server/models/executor.js
Setogit: Changed to "passing noop and setImmediate(next)" strategy for now.

kraman:
Superseded by https://github.com/strongloop/strong-mesh-models/pull/113

PR #113, opened 2015-09-29
Add DBWatcher support

kraman:
connect to strongloop-internal/scrum-nodeops#899
connect to strongloop-internal/scrum-nodeops#949

### DIFF
--- a/client/client.js
+++ b/client/client.js
@@ -147,11 +147,10 @@ function serviceList(callback) {
 }
 Client.prototype.serviceList = serviceList;
 
-function serviceDestroy(nameOrId, callback) {
+function serviceDestroy(serviceNameOrId, callback) {
   var Service = this.models.ServerService;
 
-  var q = {where: {or: [{name: nameOrId}, {id: nameOrId}]}};
-  Service.findOne(q, function(err, service) {
+  Service.findOne(_buildQuery(serviceNameOrId), function(err, service) {
     if (err || !service) return callback(err, service);
 
     Service.deleteById(service.id, callback);
@@ -170,31 +169,13 @@ Client.prototype.serviceDestroy = serviceDestroy;
  */
 function serviceFind(serviceNameOrId, callback) {
   var Service = this.models.ServerService;
-
-  var filter = {order: ['id ASC']};
-  if (serviceNameOrId) {
-    filter.where = {
-      or: [
-        {name: serviceNameOrId},
-        {id: serviceNameOrId}
-      ]
-    };
-  }
-  Service.findOne(filter, callback);
+  Service.findOne(_buildQuery(serviceNameOrId), callback);
 }
 Client.prototype.serviceFind = serviceFind;
 
 function groupCreate(serviceNameOrId, groupName, scale, callback) {
   var Service = this.models.ServerService;
-  var filter = {
-    where: {
-      or: [
-        {name: serviceNameOrId},
-        {id: serviceNameOrId}
-      ]
-    }
-  };
-  Service.findOne(filter, function(err, service) {
+  Service.findOne(_buildQuery(serviceNameOrId), function(err, service) {
     if (!service) {
       // We could create it, but then it wouldn't have a default. Until further
       // requirements, groups can only be set on existing services.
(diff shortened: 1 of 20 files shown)


client/client.js
```diff
@@ -150,8 +150,7 @@ Client.prototype.serviceList = serviceList;
 function serviceDestroy(nameOrId, callback) {
```
sam-github: should this arg be `serviceNameOrId`, or is it different from the other args with that name?

common/models/agent-trace.json
```diff
@@ -5,15 +5,18 @@
     "processId": {
       "type": "number",
       "description": "id of ServiceProcess, not the pid",
-      "required": true
+      "required": true,
+      "postgresql": {"column": "processId"}
```
sam-github: I wish we had comments :-(

What does this mean? Why doesn't it default to the name of the field?

server/models/executor.js
```diff
@@ -1,14 +1,20 @@
 var async = require('async');
+var crypto = require('crypto');
+var observerHelper = require('./observerHelper');
 
 module.exports = function(Executor) {
+  Executor.definition.properties.token.default = function() {
+    return crypto.randomBytes(24).toString('hex');
```
sam-github: https://nodejs.org/api/crypto.html#crypto_crypto_randombytes_size_callback

That's sync, it would be better to use a callback instead if that is possible here (it looks like it may not be).

server/models/gateway.js
```diff
@@ -1,13 +1,20 @@
 var async = require('async');
+var crypto = require('crypto');
+var observerHelper = require('./observerHelper');
 
 module.exports = function extendGateway(Gateway) {
-  // Refer to server-service.js for description of instance vs currentInstance
-  // vs data.
+  Gateway.definition.properties.token.default = function() {
+    return crypto.randomBytes(24).toString('hex');
```
sam-github: maybe worth putting in a util file, so at least the 24 is shared among the various default functions.

server/models/service-process.js
```diff
@@ -169,18 +170,37 @@ module.exports = function extendServiceProcess(ServiceProcess) {
...
-    return ServiceProcess.findOne.bind(ServiceProcess, filter);
+    var filter = {where: procData};
+    debug('_findOrCreateProcess: %j', procData);
+
+    // Try to find an existing process or create a new one. If an error occurs,
+    // it might be a duplicate key error, in which case, the process was already
+    // created by a parallel invocation, attempt to find that process.
+
+    // XXX(KR) The processes create does not include ppid, which may cause race
+    // conditions during cleanup due to supervisor exit.
```
sam-github: is there anything we can do to fix this? should we make sure we cleanup on start any processes which don't have a ppid, since they will never get one? Or, is that only true for the pm use-case, not central?

server/server.js
```diff
@@ -87,6 +88,20 @@ function server(serviceManager, minkelite, options) {
     }
   };
 
+  app.useDbWatcher = function(dbWather, callback) {
```
sam-github: "dbWatcher" (spelling), also below

sam-github:
PR needs squashing or rebasing, there are early commits that intro questionable code that is then replaced by later commits, so it only makes sense now when looked at as a whole. Would be nice to have a series of smaller commits, like the one that adds all the postgres cruft to the json, but if that is too hard to rebase just squashing would be preferable, IMO.

common/models/agent-trace.json
kraman: by default postgres would name this column `processid` and that messes up the mapping when it is read back into JS. Adding this column mapping fixes this.

server/models/executor.js
kraman: The default function is sync.

server/models/service-process.js
kraman: We can. Supervisor will need to be modified to send the ppid with all messages. I would prefer to handle that as a separate PR.

kraman:
PR was squashed. Sam's comments have been addressed.

sam-github:
LGTM

common/models/agent-trace.json
kraman: Related issue here: https://github.com/strongloop/loopback-connector-postgresql/issues/38

PR #114, opened 2015-10-01
Fix processs creation/query race condition

kraman:


kraman:
@slnode test please

common/models/service-process.json
```diff
@@ -99,6 +99,11 @@
       "model": "ServiceMetric"
     }
   },
+  "indexes": {
+    "uniq_process": {
+      "keys": ["pid", "workerId", "startTime", "serviceInstanceId"]
```
sam-github: this is an extra compound index that is required to be unique? and its supported in all LB connectors, or least the ones we use? I wish that json could have comments :-(

server/models/service-process.js
```diff
@@ -8,23 +8,18 @@ module.exports = function extendServiceProcess(ServiceProcess) {
...
-      debug('upsert Process: %j', err || proc, pInfo);
-      if (err) return callback(err);
-      proc.updateAttributes({
-        stopReason: '',
-        stopTime: null,
-      }, callback);
-    });
+    _findOrCreateProcess(
+      instanceId, +pInfo.wid, +pInfo.pid, +pInfo.pst || +pInfo.startTime,
+      function(err, proc) {
```
sam-github: nit: if you name this function, you won't have to wrap line 11, and you will be able to outdent the named function to its previous level, and since its body hasn't change, there won't be any diff left here, so it will be really obvious that this is a refactor to delete the direct call to .findOrCreate, and replace with a call to _findOrCreateProcess.

server/models/service-process.js
```diff
@@ -169,18 +170,37 @@ module.exports = function extendServiceProcess(ServiceProcess) {
...
-    return ServiceProcess.findOne.bind(ServiceProcess, filter);
+    var filter = {where: procData};
+    debug('_findOrCreateProcess: %j', procData);
+
+    // Try to find an existing process or create a new one. If an error occurs,
+    // it might be a duplicate key error, in which case, the process was already
+    // created by a parallel invocation, attempt to find that process.
+
+    // XXX(KR) The processes create does not include ppid, which may cause race
+    // conditions during cleanup due to supervisor exit.
```
sam-github: This is going to come back and bite us, isn't it? Can we backlog something to fix this? I assume we need to add ppid to some of the messages coming from supervisor (or all of them, like we do with pst).

sam-github:
Other than a couple very minor suggestions, and a question, LGTM.

kraman:
@slnode test please

kraman:
@slnode test please

PR #115, opened 2015-10-01
make InstanceActions ephemeral

rmg:
Instead of persisting them to disk and waiting for that persistence to
finish before handing back the result, just use the memory connector.

Furthermore, there isn't any reason to keep them around at all, so for
the immediate future we can use a simple 10s expiration so that the
instances can GC themselves instead of just accumulating in memory
indefinitely.

Connect to strongloop-internal/scrum-nodeops#908
Connect to strongloop-internal/scrum-nodeops#924
Connect to strongloop-internal/scrum-nodeops#961
Connect to strongloop-internal/scrum-nodeops#962

### Note
- This is meant as a short term solution with the eventual goal being to completely remove the InstanceAction model.
- This PR is against the 8.x branch and will need to be forward ported to master after merge.

### DIFF
--- a/server/models/instance-action.js
+++ b/server/models/instance-action.js
@@ -6,6 +6,14 @@ var path = require('path');
 var util = require('util');
 
 module.exports = function extendInstanceAction(InstanceAction) {
+  InstanceAction.observe('after save', function(ctx, next) {
+    // There is no reason to keep these instances around. Until we transition
+    // away from them entirely we'll just auto-expire them.
+    if (ctx.instance && ctx.isNewInstance) {
+      setTimeout(ctx.instance.destroy.bind(ctx.instance), 10 * 1000).unref();
+    }
+    next();
+  });
   InstanceAction.observe('before save', function(ctx, next) {
     if (!ctx.instance) return next();
     var now = Date.now();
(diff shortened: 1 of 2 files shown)


rmg:
@slnode test please

rmg:
@slnode test please

rmg:
failures are unrelated and downstream, so I'm landing this as is.

rmg:
@slnode test please

rmg:
@slnode test please

server/models/instance-action.js
```diff
@@ -6,6 +6,17 @@ var path = require('path');
 var util = require('util');
 
 module.exports = function extendInstanceAction(InstanceAction) {
+  InstanceAction.observe('after save', function(ctx, next) {
+    // There is no reason to keep these instances around. Until we transition
+    // away from them entirely we'll just auto-expire them.
+    if (ctx.instance && ctx.isNewInstance) {
+      var t = setTimeout(ctx.instance.destroy.bind(ctx.instance), 10 * 1000);
+      if (t.unref) {
```
sam-github: timeouts don't always have an unref?

sam-github:
LGTM, though I didn't know there were conditions that timers could not have an `.unref()`, what are they?

server/models/instance-action.js
rmg: I thought there was a time when they didn't.. but I suppose if there was it was pre-0.10, so I don't need this anyway.

PR #117, opened 2015-11-04
Bluemix fixes

kraman:


kraman:
@rmg not sure why pm is failing. tests seem to have been interrupted ...

rmg:
@slnode test please

rmg:
Changes LGTM. I'm also not sure why CI failed. Would be nice to figure out what the problem was, but I don't think that should block this.

PR #119, opened 2015-11-17
Change streams to track instance/process changes

kraman:
Connect to strongloop-internal/scrum-nodeops#945

### DIFF
--- a/server/boot/observe-changes.js
+++ b/server/boot/observe-changes.js
@@ -1,0 +1,52 @@
+var _ = require('lodash');
+var debug = require('debug')('strong-mesh-models:server:observe-changes');
+
+module.exports = function observeChanges(server) {
+  var ServiceInstance = server.models.ServiceInstance;
+  var ServiceProcess = server.models.ServiceProcess;
+  var ServerService = server.models.ServerService;
+
+  ServiceInstance.observe('after save', function(ctx, next) {
+    var inst = ctx.instance || ctx.data;
+    ServerService.__changeEmitter.emit('svc-' + inst.serverServiceId, {
+      target: 'ServiceInstance-' + inst.id,
+      type: ctx.isNewInstance ? 'create' : 'update',
+      data: _.pick(inst.toObject(), [
+        'serverServiceId',
+        'groupId',
+        'startTime',
+        'stopTime',
+        'restartCount',
+        'started',
+        'setSize',
+        'cpus',
+        'tracingEnabled',
+        'applicationName',
+        'npmModules',
+        'agentVersion',
+        'debuggerVersion',
+        'id',
+      ]),
+      optimistic: false,
+    });
+    setImmediate(next);
+  });
+
+  ServiceProcess.observe('after save', function(ctx, next) {
+    var proc = ctx.instance || ctx.data;
+    ServiceInstance.findById(proc.serviceInstanceId, function(err, instance) {
+      if (err) {
+        debug('Unable to locate instnce for proc %j', proc);
+        return next(err);
+      }
+      proc.serverServiceId = instance.serverServiceId;
+      ServerService.__changeEmitter.emit('svc-' + proc.serverServiceId, {
+        target: 'ServiceProcess-' + proc.id,
+        type: ctx.isNewInstance ? 'create' : 'update',
+        data: proc.toObject(),
+        optimistic: false,
+      });
+      setImmediate(next);
+    });
+  });
+};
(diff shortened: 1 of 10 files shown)


server/boot/observe-changes.js
```diff
@@ -0,0 +1,42 @@
...
+  var ServiceInstance = server.models.ServiceInstance;
+  var ServiceProcess = server.models.ServiceProcess;
+  var ServerService = server.models.ServerService;
+
+  ServiceInstance.observe('after save', function(ctx, next) {
+    var inst = ctx.instance || ctx.data;
+    ServerService.__changeEmitter.emit('svc-' + inst.serverServiceId, {
+      target: 'ServiceInstance-' + inst.id,
+      type: ctx.isNewInstance ? 'create' : 'update',
+      data: _.omit(inst.toObject(), [
```
rmg: Would it be better to take a whitelist approach here instead and list only the fields that we _do_ want? That way in the future if a sensitive field is added to the model we can't accidentally forget to hide it here.

server/boot/observe-changes.js
```diff
@@ -0,0 +1,42 @@
...
+      ]),
+      optimistic: false,
+    });
+    setImmediate(next);
+  });
+
+  ServiceProcess.observe('after save', function(ctx, next) {
+    var proc = ctx.instance || ctx.data;
+    ServiceInstance.findById(proc.serviceInstanceId, function(err, instance) {
+      if (err) return debug('Unable to locate instnce for proc %j', proc);
```
rmg: Does the observers API not expect `next()` to be called with an error in this case?

server/models/server-service.js
```diff
@@ -249,4 +253,55 @@ module.exports = function extendServerService(ServerService) {
...
+
+    changes.on('error', function() {
+      changes.destroy();
+    });
+
+    changes.on('end', function() {
+      changes.destroy();
+    });
+
+    function publishChange(data) {
```
rmg: Is this intentionally not at the beginning or end of the function?

test/test-notification-model-updates.js
```diff
@@ -519,6 +534,24 @@ test('Test notifications', function(t) {
...
+  t.test('check change stream', function(tt) {
+    var change = changeStream.read();
+    tt.equal(change.target, 'ServiceInstance-1');
+    tt.equal(change.type, 'create');
+    change = changeStream.read();
+    tt.equal(change.target, 'ServiceInstance-1');
+    tt.equal(change.type, 'update');
+    change = changeStream.read();
+    var eventCnt = 0;
+    while (change !== null) {
```
rmg: Is this being done for debug reasons? I expected something more like:
```js
while (changeStream.read() !== null) {
  eventCnt++;
}
```

I'm mostly wondering if I'm missing something..

test/test-notification-model-updates.js
```diff
@@ -519,6 +534,24 @@ test('Test notifications', function(t) {
...
+      eventCnt++;
+      change = changeStream.read();
+    }
+    tt.equal(eventCnt, 21);
```
rmg: What is the significance of this `21`? An assertion description would be particular appropriate here so that if the tests fail at some point because there are 20 or 22 someone will have a hope of figuring out what that means.

server/models/server-service.js
```diff
@@ -249,4 +253,55 @@ module.exports = function extendServerService(ServerService) {
...
+    changes._serviceId = this.id;
+    debug('change stream %s created', changes._serviceId);
+    ServerService.__changeEmitter.on(
+      'svc-' + changes._serviceId,
+      publishChange
+    );
+
+    changes.destroy = function() {
+      debug('change stream %s destroyed', changes._serviceId);
+      changes.removeAllListeners('error');
```
rmg: Should this be removing all listeners instead of just these 2? Seems to me like it could theoretically still be emitting `data` events to someone holding on to a ref.. but if it tried to emit an error as a result, nothing would be listening so it would crash the app.

server/models/server-service.js
```diff
@@ -1,3 +1,5 @@
+var PassThrough = require('stream').PassThrough;
```
rmg: Wrong order ;-)

rmg:
A couple nits and a couple concerns/questions. Back to your, @kraman.

server/boot/observe-changes.js
kraman: good catch. fixed

server/models/server-service.js
kraman: moved to end

test/test-notification-model-updates.js
kraman: yah, i have a console.log in there earlier. i simplified it.

test/test-notification-model-updates.js
kraman: added comment

rmg:
LGTM.

PR #120, opened 2015-12-31
Make tests pass on Windows

rmg:
* Use smarter assertions (`t.match(a, b)` vs `t.assert(b.test(a))`) for improved test debugging.
* Avoid `process.exit()` where possible to ensure stdio is flushed, allowing expected error messages to be seen
* Skip SSH based tests on Windows since it is unlikely to have sshd running locally

### DIFF
--- a/bin/sl-meshctl.js
+++ b/bin/sl-meshctl.js
@@ -46,12 +46,10 @@ while ((option = parser.getopt()) !== undefined) {
   switch (option.option) {
     case 'v':
       console.log(require('../package.json').version);
-      process.exit(0);
-      break;
+      return;
     case 'h':
       printHelp($0, console.log);
-      process.exit(0);
-      break;
+      return;
     case 'C':
       apiUrl = option.optarg;
       break;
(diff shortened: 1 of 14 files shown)


rmg:
@slnode test please

rmg:
@slnode test please

rmg:
@slnode test please

rmg:
@slnode test please

bin/sl-meshctl.js
```diff
@@ -92,11 +90,11 @@ if (process.env.SSH_PORT) {
 dieIf.url = apiUrl;
 maybeTunnel(apiUrl, sshOpts, function(err, apiUrl) {
   dieIf.url = apiUrl || dieIf.url;
-  dieIf(err);
+  if (err) return dieIf(err);
```
sam-github: https://github.com/strongloop/strong-mesh-models/blob/cleanup-tests/bin/sl-meshctl.js#L827

This reads funny "if error die if error"... The "if" is what `dieIf()` is already supposed to do. I see your problem, you want to return early. `dieIf()` should renamed `die()` if the `if` always has to be done outside the function.

sam-github:
LGTM

bin/sl-meshctl.js
rmg: Agreed. I was trying to keep the changes minimal, but I probably overshot.

rmg:
@slnode test please

